Issue #361, opened 2014-11-13
Assignment Finalization


vishalkal:
On the assignments page, have a checkbox on the side of a row indicating whether or not an advisor wants to keep or remove an assignment. At the bottom, create a button that, upon being clicked, will bring up an agreement stating that this will finalize the assignments they want to keep and permanently give up all the rest.

nathanielparke:
completed via 0c5fe7d03f96a280b1ecdb2dd64f826392bf3ed8

PR #376, opened 2014-12-22
issue #361. Draft code for assignment finalization.

patilpranay:


huxley/api/views/school.py
```diff
@@ -41,6 +40,13 @@ def get_queryset(self):
 
         return Assignment.objects.filter(school_id=school_id)
 
+class SchoolFinalize(generics.CreateAPIView):
```
kmeht: Doesn't make much sense to use a CreateAPIView here, since that's specifically designed for creating resources, e.g. models. Instead, you can just extend from `GenericAPIView` (I believe) and leave everything else the same.

huxley/api/urls.py
```diff
@@ -27,7 +27,8 @@
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
-    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice')
+    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice'),
+    url(r'^schools/(?P<pk>[0-9]+)/finalize/?$', views.school.SchoolFinalize.as_view(), name='school_finalize')
```
kmeht: Add a trailing comma at the end of this URI declaration, so we don't have to modify the line when more are added.

huxley/core/models.py
```diff
@@ -118,6 +118,8 @@ class School(models.Model):
     fees_owed = models.DecimalField(max_digits=6, decimal_places=2, default=0)
     fees_paid = models.DecimalField(max_digits=6, decimal_places=2, default=0)
 
+#    assignments_finalized = models.BooleanField(default=False)
```
kmeht: Remember that you'll need to create a migration when you add this field.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -68,12 +69,59 @@ var AdvisorAssignmentsView = React.createClass({
               {this.renderAssignmentRows()}
             </table>
           </div>
-          <div className="tablemenu footer" />
+          <div className="tablemenu footer">
+            <Button
+              color="green"
+              size="small"
+              loading={this.state.loading}
```
kmeht: Add a default value to `getInitialState` for this property.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -68,12 +69,59 @@ var AdvisorAssignmentsView = React.createClass({
...
         </form>
       </InnerView>
     );
   },
 
+  _handleClick: function(event) {
+    this.setState({loading: true});
+    var user = this.props.user.getData();
+    var school = this.props.user.getSchool();
+    var elements = document.getElementsByTagName("input");
```
kmeht: The whole point of React is to abstract the DOM away. Don't do this. Instead, whatever data you want to send should be inside this component's `state`. A component's props and state are its source of truth: the `render` function uses that state to build the UI, and if you need to send data to the server, that's where it comes from.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -68,12 +69,59 @@ var AdvisorAssignmentsView = React.createClass({
...
+    for (var i=0; i<elements.length; i++) {
+      if (elements[i].checked) {
+        assignList.push(elements[i].getAttribute("name"));
+      }
+    }
+    window.alert("worked");
+    $.ajax({
+      type: 'POST',
+      url: '/api/schools/' + school.id + '/finalize/',
+      data: assignList,
```
kmeht: The data should JSON-encoded (If I'm not mistaken). Check out other requests and see if that's the case.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -83,6 +131,7 @@ var AdvisorAssignmentsView = React.createClass({
           <td>{committees[assignment.committee].name}</td>
           <td>{countries[assignment.country].name}</td>
           <td>{committees[assignment.committee].delegation_size}</td>
+          <td><input name={assignment.id} type="checkbox"/></td>
```
kmeht: Currently these are uncontrolled components. Instead, they should be controlled components, so that their values can be encoded in state. Use ReactLink to do this: http://facebook.github.io/react/docs/two-way-binding-helpers.html

@tdowds has used it (relatively) extensively and should be able to help you.

huxley/core/migrations/0001_initial.py
```diff
@@ -252,8 +252,8 @@ def backwards(self, orm):
             'state': ('django.db.models.fields.CharField', [], {'max_length': '16'}),
             'times_attended': ('django.db.models.fields.PositiveSmallIntegerField', [], {'default': '0'}),
             'waitlist': ('django.db.models.fields.BooleanField', [], {'default': 'False'}),
-            'zip_code': ('django.db.models.fields.CharField', [], {'max_length': '16'})
+            'zip_code': ('django.db.models.fields.CharField', [], {'max_length': '16'}),
```
vishalkal: Looks like the assignments_finalized attribute hasn't been migrated so the Travis build is failing.

kmeht: @patilpranay, there should be a generated migration attached to this diff that adds the `assignments_finalized` field. You may have generated it but forgotten to add it to the commit?

huxley/api/urls.py
```diff
@@ -27,7 +27,8 @@
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
-    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice')
+    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice'),
+    url(r'^schools/(?P<pk>[0-9]+)/finalize/?$', views.school.SchoolFinalize.as_view(), name='school_finalize'),
```
kmeht: Bit of feedback, I think the URI should be `/school/:id/assignments/finalize`, and the view should be `SchoolAssignmentsFinalize`.

kmeht:
Closing an old PR.

PR #378, opened 2015-01-13
Regclose

tdowds:
Closed registration on the backend. Created a setting that was true if registration is open. When registration is closed set the setting to false which raises a PermissionDenied error in huxley/api/views/user.py when trying to create a user. Updated the unit tests accordingly by overriding the setting, and also created a unit test to test for closed registration.

### DIFF
--- a/huxley/api/tests/test_user.py
+++ b/huxley/api/tests/test_user.py
@@ -254,6 +254,7 @@ class UserListGetTestCase(ListAPITestCase):
              'school': user2.school_id,
              'committee': user2.committee_id}])
 
+
 @override_settings(REGISTRATION_OPEN=True)
 class UserListPostTestCase(CreateAPITestCase):
     url_name = 'api:user_list'
--- a/huxley/api/tests/test_user.py
+++ b/huxley/api/tests/test_user.py
@@ -6,6 +6,7 @@ import json
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test.client import Client
+from django.test.utils import override_settings
 
 from huxley.accounts.models import User
 from huxley.api.tests import (CreateAPITestCase, DestroyAPITestCase,
@@ -253,7 +254,7 @@ class UserListGetTestCase(ListAPITestCase):
              'school': user2.school_id,
              'committee': user2.committee_id}])
 
-
+@override_settings(REGISTRATION_OPEN=True)
 class UserListPostTestCase(CreateAPITestCase):
     url_name = 'api:user_list'
     params = {'username': 'Kunal',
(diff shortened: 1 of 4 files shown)


huxley/api/views/user.py
```diff
@@ -24,7 +24,9 @@ class UserList(generics.ListCreateAPIView):
 
     def get_serializer_class(self):
         if self.request.method == 'POST':
-            return CreateUserSerializer
+            if settings.REGISTRATION_OPEN:
```
kmeht: I'm not convinced that this is the best spot to put this. Instead, we should override the `create` method on this view, and either raise PermissionDenied or call the parent method based on the setting.

huxley/settings/registration.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2011-2014 Berkeley Model United Nations. All rights reserved.
+# Use of this source code is governed by a BSD License (see LICENSE).
+
+REGISTRATION_OPEN = False
```
kmeht: You'll also have to import this setting into the __init__.py file of this directory, otherwise it won't be available in `django.conf.settings`.

huxley/api/views/user.py
```diff
@@ -16,7 +17,6 @@
 from huxley.api.serializers import CreateUserSerializer, UserSerializer
 from huxley.core.models import School
 
-
```
kmeht: Please restore this line. (Two lines after imports)

huxley/api/tests/test_user.py
```diff
@@ -253,7 +254,7 @@ def test_superuser(self):
              'school': user2.school_id,
              'committee': user2.committee_id}])
 
-
+@override_settings(REGISTRATION_OPEN=True)
```
kmeht: Please add a line above this (two lines between top-level blocks)

tdowds:
Overrided the create method.

huxley/api/views/user.py
```diff
@@ -22,6 +23,11 @@ class UserList(generics.ListCreateAPIView):
     queryset = User.objects.all()
     permission_classes = (IsPostOrSuperuserOnly,)
 
+    def create(self, request, *args, **kwargs):
+        if settings.REGISTRATION_OPEN:
+            return super(UserList, self).create(request, args, kwargs)
```
kmeht: You have to explode the args and kwargs when you call the super method, otherwise they're passed in as just regular params as a tuple and dictionary, respectively.
```python
return super(UserList, self).create(request, *args, **kwargs)
```

tdowds:
updated

Issue #46, opened 2013-06-12
Add Logging


kmeht:
Develop a system of logging to be used throughout the app. I am way over seeing a 500 internal server error and then bending over backward to debug it. We should write clear, concise, descriptive logs throughout the app's execution.

Subtasks:
- [x] Add a basic console logger
- [x] Log exceptions
- [x] Add log statements tracking the lifecycle of requests
- [x] Create a `logging` app with a `LogEntry` model
- [x] Set up log settings to log to console in dev, and log to `logging` in stage/production

tdowds:
closed via 3d94ccbb26c72c4a77db14505c713ab2bf329504

PR #379, opened 2015-01-16
#46 middleware for handling exceptions.

tdowds:
Pushed this to see if this was a good start in regards to exception handling. For some reason the logger won't work in the ExceptionLoggerMiddleware class. It fails on the line 'logger = logging.getLogger('huxley')'. However the ExceptionLoggerMiddleware class's process_exception method does get run (I put a print statement in to test, and forced an exception in one of the views.)


kmeht:
Is there an exception or something (hah, irony) thrown in there? If so, mind pastebinning it?

tdowds:
renamed huxley/settings/middlewares.py to huxley/core/middlewares.py

PR #383, opened 2015-01-29
Includes functionality to upload delegates in admin panel.

vishalkal:


huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,25 @@ def roster(self, request):
 
         return roster
 
+    def load(self, request):
+        '''Loads new Assignments.'''
+        delegates = request.FILES
+        reader = csv.reader(delegates['csv'])
+
+        def upload_delegates(reader):
```
kmeht: Why is this a separate function? It's only used once, so we can just inline it.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,25 @@ def roster(self, request):
...
+            assignments = {}
+            header = False
+            for row in reader:
+                if not header:
+                        header = True
```
kmeht: These two lines are double-indented.

kmeht:
In general, I think this is a pretty crazily inefficient algorithm. Each `Foo.objects.get` is a separate call to the database, unless I'm mistaken. Instead, I think it would be a better idea to just get all the objects for each particular model, and then construct a dictionary from name to ID. That way, the number of trips to DB would be significantly reduced.

kmeht:
What if you have to re-upload the delegates file? This code will just keep creating new delegates, and duplicating the ones that are already there. Look at how the assignments upload algorithm does it. It checks if the delegate is already there and updates it, otherwise creates a new one.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,25 @@ def roster(self, request):
...
         return roster
 
+    def load(self, request):
+        '''Loads new Assignments.'''
+        delegates = request.FILES
+        reader = csv.reader(delegates['csv'])
+
+        assignments = {}
+        for assignment in Assignments.objects.all():
+            assignments[committee.name + country.name + school.name] = assignment
```
kmeht: Where are `committee`, `country`, and `school` defined? Did you even test this code?

vishalkal: Weird I thought I did but I must have changed the code right before I committed

huxley/core/tests/admin/test_delegate.py
```diff
@@ -0,0 +1,109 @@
...
+
+        with closing(f) as f:
+            self.client.post(reverse('admin:core_delegate_load'), {'csv': f})
+
+        self.assertTrue(Delegate.objects.filter(
+            assignment=Assignment.objects.get(
+                school=School.objects.get(name='Test School'),
+                committee=Committee.objects.get(name='SPD'),
+                country=Country.objects.get(name="Côte d'Ivoire")
+                ),
```
kmeht: Outdent this paren one level. Same for the assertion below.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,23 @@ def roster(self, request):
...
         return roster
 
+    def load(self, request):
+        '''Loads new Assignments.'''
+        delegates = request.FILES
+        reader = csv.reader(delegates['csv'])
+
+        assignments = {}
+        for assignment in Assignment.objects.all():
+            assignments[assignment.committee.name.encode('ascii', 'ignore') + assignment.country.name.encode('ascii', 'ignore') + assignment.school.name] = assignment
```
kmeht: The string concatenation is pretty gross. Dictionaries can be keyed by tuples, let's just do that:
```python
assignments[
    assignments[assignment.committee.name.encode('ascii', 'ignore'),
    ...,
    ..., # Don't forget the trailing comma!
] = assignment
```

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,23 @@ def roster(self, request):
...
+        for row in reader:
+            if row[1] == 'Committee':
+                continue
+            assignment = assignments[unicode(row[1], errors='ignore')+ unicode(row[2], errors='ignore') + row[3]]
```
kmeht: As mentioned above, we can use tuples instead:
```python
assignment = assignments[
    unicode(row[1], errors='ignore'),
    ...,
    ..., # Don't forget the trailing comma!
]
```

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,23 @@ def roster(self, request):
...
+            d = Delegate(name=row[0], assignment=assignment)
```
kmeht: `Delegate.objects.create(...)`

huxley/core/tests/admin/test_delegate.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class DelegateAdminTest(TestCase):
+
+    def test_import(self):
+        '''Test that the admin panel can import delegates. '''
+        TestUsers.new_superuser(username='testuser', password='test')
+        self.client.login(username='testuser', password='test')
+
+        f = TestFiles.new_csv([
```
kmeht: Why is this still here? Doesn't look like it's used.

huxley/core/tests/admin/test_delegate.py
```diff
@@ -0,0 +1,64 @@
...
+            ['SPD', 'Special Pôlitical and Decolonization', 2, ''],
+            ['USS', 'United States Senate', 2, True]
+        ])
+
+        f = TestFiles.new_csv([
```
kmeht: Same here, doesn't look like it's used.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,29 @@ def roster(self, request):
...
+             ] = assignment
+        for row in reader:
+            if row[1] == 'Committee':
+                continue
+            assignment = assignments[unicode(row[1], errors='ignore'),
+                                     unicode(row[2], errors='ignore'),
+                                     row[3], 
+                                    ]
+            d = Delegate.objects.create(name=row[0], assignment=assignment)
+            d.save()
```
kmeht: `create` is a shortcut for constructing and saving a model. This is a redundant call.

huxley/core/admin/delegate.py
```diff
@@ -5,9 +5,10 @@
 
 from django.conf.urls import patterns, url
 from django.contrib import admin
-from django.http import HttpResponse
+from django.core.urlresolvers import reverse
+from django.http import HttpResponse, HttpResponseRedirect
 
-from huxley.core.models import Delegate
+from huxley.core.models import Assignment, Committee, Country, Delegate, School
```
kmeht: Doesn't look like the Committee, Country, and School classes are used in this file.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,29 @@ def roster(self, request):
...
         return roster
 
+    def load(self, request):
+        '''Loads new Assignments.'''
+        delegates = request.FILES
+        reader = csv.reader(delegates['csv'])
+
+        assignments = {}
+        for assignment in Assignment.objects.all():
+            assignments[assignment.committee.name.encode('ascii', 'ignore'),
```
kmeht: This code style is kinda weird and inconsistent.
```python
assignments[
    foo,
    bar,
    baz,
] = assignment
```

This way the indentation is consistent with the meaning of each expression.

huxley/core/admin/delegate.py
```diff
@@ -29,6 +30,29 @@ def roster(self, request):
...
+             assignment.country.name.encode('ascii', 'ignore'),
+             assignment.school.name, 
+             ] = assignment
+        for row in reader:
+            if row[1] == 'Committee':
+                continue
+            assignment = assignments[unicode(row[1], errors='ignore'),
```
kmeht: Same comment as above.

PR #384, opened 2015-02-01
#46 added file and admin_mail handlers.

tdowds:
Writes to the same file every time an exception occurs so will need to fix that in a separate diff, or at least look into more. Updated .gitignore

### DIFF
--- a/huxley/logs/handlers.py
+++ b/huxley/logs/handlers.py
@@ -13,9 +13,10 @@ class DatabaseHandler(logging.Handler):
 
     def emit(self, record):
         try:
-            logEntry = LogEntry(level=record.levelname, message=record.message, timestamp=datetime.datetime.now())
-            logEntry.save()
+            log_entry = LogEntry(
+                level=record.levelname,
+                message=record.message,
+                timestamp=datetime.datetime.now())
+            log_entry.save()
         except:
             pass
-
-        return
--- a/huxley/logs/handlers.py
+++ b/huxley/logs/handlers.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+# Use of this source code is governed by a BSD License (see LICENSE).
+
+import datetime
+import logging
+
+from huxley.logs.models import LogEntry
+
+
+class DatabaseHandler(logging.Handler):
+    def __init__(self):
+        logging.Handler.__init__(self)
+
+    def emit(self, record):
+        try:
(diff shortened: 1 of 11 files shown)


kmeht:
Hm, according to the Travis build failure we have to manually check if the file is there... it doesn't automatically create it for us? That's a bummer :(

huxley/settings/logging.py
```diff
@@ -15,6 +15,16 @@
             'class': 'logging.StreamHandler',
             'formatter': 'simple',
         },
+        'file': {
+            'level': 'DEBUG',
+            'class': 'logging.FileHandler',
+            'filename': 'huxley/logs/error.log',
```
kmeht: If this handler is for levels DEBUG and above, I don't think the filename should be named `error`.

tdowds:
updated

huxley/logs/handlers.py
```diff
@@ -0,0 +1,21 @@
...
+        logging.Handler.__init__(self)
+
+    def emit(self, record):
+        try:
+            logEntry = LogEntry(level=record.levelname, message=record.message, timestamp=datetime.datetime.now())
+            logEntry.save()
+        except:
+            pass
+
+        return
```
kmeht: You don't need this explicit return, since you're just ending the function.

huxley/logs/handlers.py
```diff
@@ -0,0 +1,21 @@
...
+from huxley.logs.models import LogEntry
+
+
+class DatabaseHandler(logging.Handler):
+    def __init__(self):
+        logging.Handler.__init__(self)
+
+    def emit(self, record):
+        try:
+            logEntry = LogEntry(level=record.levelname, message=record.message, timestamp=datetime.datetime.now())
```
kmeht: `log_entry`, convention is underscores in Python.

kmeht: Also, cap lines at 80 characters. This would become something like:
```python
log_entry = LogEntry(
    level=record.levelname,
    message=record.message,
    timestamp=...)
```

huxley/logs/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
kmeht: Delete this file until we need it.

huxley/logs/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
kmeht: Delete this file until we need it.

tdowds:
updated

Issue #374, opened 2014-12-09
School Model Fee Calculation


vishalkal:
Change the school model to use assignments, and not expected number of delegates.

tdowds:
Closed due to change in preferences from treasurer and SG

PR #385, opened 2015-02-06
issue #374. Changed how fees are calculated.

patilpranay:


huxley/core/models.py
```diff
@@ -148,11 +148,7 @@ def update_country_preferences(self, country_ids):
     @classmethod
     def update_fees(cls, **kwargs):
         school = kwargs['instance']
-        delegate_fees = cls.DELEGATE_FEE * sum((
-            school.beginner_delegates,
-            school.intermediate_delegates,
-            school.advanced_delegates,
-        ))
+        delegate_fees = cls.DELEGATE_FEE * len(Assignment.objects.filter(school_id=school.id))
```
kmeht: You don't need to fetch all the assignments to count them. Check out the QuerySet API and see if you can find a better function. (Spoiler alert: there's totally a better function).

https://docs.djangoproject.com/en/1.6/ref/models/querysets/

huxley/core/models.py
```diff
@@ -180,6 +176,7 @@ def save(self, *args, **kwargs):
         if getattr(self, '_pending_country_preference_ids', []):
             self.update_country_preferences(self._pending_country_preference_ids)
             self._pending_country_preference_ids = []
+            post_save.connect(School.update_fees, sender=School)
```
kmeht: Why do we need to add this here? We've already hooked up `update_fees` to the `pre_save` signal below... why do we need to calculate it twice?

huxley/core/tests/models/test_models.py
```diff
@@ -92,27 +92,28 @@ def test_update_country_preferences(self):
...
+            Assignment(committee_id=cm, country_id=ct, school_id=school.id)
+            for ct in [ct1, ct2, ct3]
+            for cm in [cm1, cm2]
+        ])
         school.save()
 
         self.assertEquals(
             school.fees_owed,
-            school.REGISTRATION_FEE + school.DELEGATE_FEE * (b2 + i2 + a2),
+            school.REGISTRATION_FEE + school.DELEGATE_FEE * 6,
```
kmeht: Shouldn't we take into account the number of delegates per delegation? Or are the fees calculated per-assignment instead of per-delegate?

kmeht:
@patilpranay, @vishalkal, what's going on with this PR?

patilpranay:
Sorry I'll get it out today evening.

nathanielparke:
Updates on this?

kmeht:
Closing old PR.

Issue #386, opened 2015-03-20
Update to Django 1.7


kmeht:
Django 1.7 introduces a few new features, the biggest one being native database migrations. This will break our dependency on South! It also has a few new features for initializing apps, which may be helpful to us.

nathanielparke:
We should probably get this done before #331 so we can manage the new database through Django

PR #389, opened 2015-05-14
#386 django1.7

nathanielparke:
Updates to django 1.7 and breaks south dependency. The previous migrations were deleted but the directory and the __init__.py files were kept. Then "python manage.py makemigrations" which created a new initial migration. Then those migrations were faked with "python manage.py mirgrate --fake APPLICATION" where APPLICATION was first core then payments. The fake tag was necessary because the changes had already been applied to the db so django internally just needed to mark the tables as applied. More information can be found at https://docs.djangoproject.com/en/1.8/topics/migrations/#upgrading-from-south or https://docs.djangoproject.com/en/1.8/releases/1.7/


kmeht:
Cool! Could you update the description to indicate what steps were taken with the migrations? Were they all collapsed into one Django migration?




A link to the documentation would also be helpful. 



—

PR #392, opened 2015-05-28
#46 log tests round 1.

tdowds:
Definitely more tests to go, just starting simple for now.

### DIFF
--- a/huxley/logs/handlers.py
+++ b/huxley/logs/handlers.py
@@ -13,10 +13,11 @@ class DatabaseHandler(logging.Handler):
 
     def emit(self, record):
         try:
+            self.format(record)
             log_entry = LogEntry(
                 level=record.levelname,
-                message=record.msg,
-                timestamp=datetime.datetime.now())
+                message=record.message,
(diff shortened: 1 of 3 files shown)


kmeht:
Good start, but this doesn't really test any application functionality. Let's add a few more tests before merging.

tdowds:
Added a test to test the DatabaseHandler. I adjusted a couple things in the handler. First I added self.format(record) because it allows access to the message attribute of the LogRecord object being passed to the emit function. Also I set timestamp to None because currently our formatter for the DatabaseHandler doesn't include a timestamp (%(asctime)s) in the LogRecord object being passed to the DatabaseHandler's emit method. I was running into this problem when writing the test because I was trying to compare the timestamp of the LogRecord object I created in the test, log_record, to the LogEntry object created by the DatabaseHandler, log_entry, but log_record didn't have any attribute ('asctime') to access its timestamp because it's not formatted to do so. So I decided to stick with no time stamp for now, and then potentially update the formatter in logging.py in the next diff if that's what we decide to do.

Issue #393, opened 2015-05-30
Create List of Specialized Committees 


nathanielparke:
In the registration view we currently have a section that asks advisors if they would be interested in participating in specialized committees but we do not allow them to choose which specialized committees they want to send delegates to. We should change this so that registration has a list of specialized committees to choose from.

kmeht:
We totally used to have this. Guess we removed it for 63?

tdowds:
Is there more to do on this?

kmeht:
@tdowds, do the list of specialized committees appear on the registration page? That would be the UI portion of this task.

tdowds:
you mean this?
<img width="499" alt="screen shot 2015-07-16 at 9 39 50 pm" src="https://cloud.githubusercontent.com/assets/5726560/8740727/3c87eef0-2c03-11e5-9338-6c09edf72c95.png">

kmeht:
Oh, great! So if registration works correctly (i.e. those preferences are saved), you can close this out.

PR #396, opened 2015-06-11
#393 added model field and adjusted unit tests.

tdowds:
Removed the old committee preferences fields and added a ManyToManyField for committeepreferences. Adjusted unit tests accordingly. Still to do: update school serializer, update unit tests, (update Admin too?)


tdowds:
Updated the serializer and updated the unit tests

kmeht:
Looks good, but you'd still have to include the migration otherwise Travis will fail.

tdowds:
Here's the migration

kmeht:
@tdowds: Merged this PR, but Python convention is to separate words in a variable with an underscore. Should've been `committee_preferences`.

PR #398, opened 2015-07-03
updated registration ui to select committees preferences 

tdowds:
Can select your specialized committee preferences and submit them. There needs to be some work done on the admin side though to display these preferences, because all the committees show up and the ones that were selected are highlighted until you focus on them.

### DIFF
--- a/huxley/www/static/js/huxley/components/RegistrationView.js
+++ b/huxley/www/static/js/huxley/components/RegistrationView.js
@@ -406,13 +406,13 @@ var RegistrationView = React.createClass({
       return (
         <li>
           <label name="committee_prefs">
-          <input
-            className="choice"
-            type="checkbox"
-            name="committee_prefs"
-            onChange={this._handleCommitteePreferenceChange.bind(this, committee)}
-          />
-          {committee.full_name}
+            <input
+              className="choice"
+              type="checkbox"
+              name="committee_prefs"
+              onChange={this._handleCommitteePreferenceChange.bind(this, committee)}
+            />
+            {committee.full_name}
           </label>
         </li>
       );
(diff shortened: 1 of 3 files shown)


huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -89,6 +87,9 @@ var RegistrationView = React.createClass({
     CountryStore.getCountries(function(countries) {
       this.setState({countries: countries});
     }.bind(this));
+    CommitteeStore.getSpecialCommittees(function(committees) {
+      this.setState({committees: committees});
+    }.bind(this));
   },
```
nathanielparke: Variable choice here seems a bit confusing. Maybe we should consider refactoring both the arguments of the inner functions to be a little more clear as to what is going on in this function.

nathanielparke:
can you post a screenshot?

tdowds:
screen shot 2015-07-02 at 7 26 06 pm

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -449,6 +401,24 @@ var RegistrationView = React.createClass({
     );
   },
 
+  renderCommittees: function() {
+    return this.state.committees.map(function(committee) {
+      return (
+        <li>
+          <label name="committee_prefs">
+          <input
```
kmeht: Indent this entire tag one level:
```
<label name="...">
  <input
    ...
  />
  {committee.full_name}
</label>
```

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -571,6 +541,15 @@ var RegistrationView = React.createClass({
     this.setState(change);
   },
 
+   _handleCommitteePreferenceChange: function(committee) {
+    var index = this.state.committee_prefs.indexOf(committee.id);
+    if (index < 0) {
+      this.state.committee_prefs.push(committee.id);
```
kmeht: You cannot modify `this.state` directly. Changes to state have to happen through `setState`.
This, for example, would be
```js
this.setState({
  committee_prefs: this.state.committee_prefs.concat(committee.id),
});
```
We use `concat()` so that a new array is created and the old one isn't modified.

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -571,6 +541,15 @@ var RegistrationView = React.createClass({
...
+    } else {
+      this.state.committee_prefs.splice(index, 1);
```
kmeht: Similarly, here we'd use a `filter()` call:
```js
this.setState({
  committee_prefs: this.state.committee_prefs.filter(function(id) {
    return committee.id !== id;
  }),
});
```
Filtering creates a new array, and the old one isn't modified.

tdowds:
Updated

PR #399, opened 2015-07-10
#46 Changed represenation of LogEntry instance.

tdowds:
So the logs are more readable in the admin panel. Here's what it looks like
screen shot 2015-07-09 at 9 45 08 pm

### DIFF
--- a/huxley/logs/models.py
+++ b/huxley/logs/models.py
@@ -7,3 +7,6 @@ class LogEntry(models.Model):
     level = models.CharField(max_length=200)
     message = models.TextField()
     timestamp = models.DateTimeField('timestamp', null=True, blank=True)
+
+    def __str__(self):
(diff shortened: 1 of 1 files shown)


huxley/logs/models.py
```diff
@@ -7,3 +7,6 @@ class LogEntry(models.Model):
     level = models.CharField(max_length=200)
     message = models.TextField()
     timestamp = models.DateTimeField('timestamp', null=True, blank=True)
+
+    def __str__(self):
```
kmeht: If this is a unicode string, shouldn't this be the `__unicode__` method?

tdowds:
updated

kmeht:
Merged with https://github.com/bmun/huxley/commit/4c75b62c5cd7f640ce06ea69602931e62c1bc979.

PR #400, opened 2015-07-13
initial logging for log in request commit.

tdowds:
Not exactly sure what tracking the lifecycle of a request looks like in terms of logging, but I thought this would be a good place to start.

### DIFF
--- a/huxley/api/views/user.py
+++ b/huxley/api/views/user.py
@@ -17,6 +17,8 @@ from huxley.api.permissions import IsPostOrSuperuserOnly, IsUserOrSuperuser
 from huxley.api.serializers import CreateUserSerializer, UserSerializer
 from huxley.core.models import School
 
+import logging
+
 
 class UserList(generics.ListCreateAPIView):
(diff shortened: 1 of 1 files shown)


kmeht:
I'm not sure of the best route to take either. I think these will lead to a ton of extraneous logs, however, since we're adding a logging statement every step of the way. I think the most important things to log are unusual events in the program's execution.

My intuition is that a good place to start logging would be logging every response from the API, including its request URI, status code and JSON payload. For example, this will cover the cases when authentication failed and another user is logged in. It will also cover the case of a successful login.

I'd guess that this would be either a middleware, or within one of the base API views in `huxley.api`. Perhaps we have to create a APIView class that basically intercepts the response and logs it before returning it. This way we get all the API responses, but not the other crap. Or, if we went with a middleware, perhaps we could check if the request URI belongs to the `api` app, and only then log the response.

kmeht:
@tdowds, closing this out since we're taking a different approach.

PR #401, opened 2015-07-14
Added script to erase old LogEntries from the database.

tdowds:
for #46


huxley/logs/management/commands/erase_logs.py
```diff
@@ -0,0 +1,27 @@
...
+class Command(BaseCommand):
+    args = '<days> (optional).'
+    help = 'Deletes all log entries after a specified number of days (defaults to 30).'
+
+    def handle(self, *args, **kwargs):
+        if args:
+            try:
+                days = int(args[0])
+            except ValueError as e:
+                raise e
```
kmeht: If we're just gonna raise the error again, may as well not even `try/except` it. Just let it happen.
```python
days = int(args[0]) if args else 30
```

PR #403, opened 2015-07-26
Make 'fab.pr' work for any upstream alias.

kmeht:
This defaults the upstream alias to 'upstream', but allows it to be configured by the .fabricrc file. This way, folks who check out the repo directly (or alias it as something else) can still use the command.

### DIFF
--- a/fabfile/pr.py
+++ b/fabfile/pr.py
@@ -4,7 +4,7 @@
 import json
 import urllib2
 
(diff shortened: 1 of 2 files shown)


nathanielparke:
Looks good to me!

kmeht:
Thanks, sorry I forgot to push this myself!



—

PR #412, opened 2015-09-14
This preps Huxley for BMUN 64

nathanielparke:


kmeht:
Using the merge button instead of fab pr?



—

nathanielparke:
So I actually did use fab PR. I'm unsure what happened here. I switched to my local master branch and ran fab PR. It created the squashed commit but ended up merging this PR instead.

kmeht:
If it's just one commit, github detects it and marks the PR as merged. If that's the case, my mistake :)



—

nathanielparke:
Yeah it was only one commit. Cool!

PR #418, opened 2015-09-21
initial invoice generator.

tdowds:
Here's what I have so far for the invoice.
A couple things still do: 
1) Encoding the logo onto the pdf: I was running into an error that said something like "NoneType" doesn't have method or attribute "encode"
2) Invoice number: How do we generate an invoice number?
3) I removed the InvoiceButton component in the AdvisorProfileView and so now the button to generate the invoice is a link, and it doesn't look very good
4) there are some extraneous commits that I need to weed out of here as well, must've came up when I resolving merge conflicts.
screen shot 2015-09-21 at 2 03 57 pm
screen shot 2015-09-21 at 2 04 18 pm


huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -9,7 +9,7 @@ var React = require('react/addons');
 
 var Button = require('./Button');
 var InnerView = require('./InnerView');
-var InvoiceButton = require('./InvoiceButton');
+//var InvoiceButton = require('./InvoiceButton');
```
kmeht: Remove this if it's not being used.

tdowds:
screen shot 2015-09-21 at 7 29 02 pm

requirements.txt
```diff
@@ -15,3 +15,6 @@ pycrypto==2.6.1
 requests==2.4.3
 six==1.7.3
 wsgiref==0.1.2
+django-easy-pdf==0.1.0
```
kmeht: It's better to dump the requirements into the requirements file than modify them directly, because the requirements may have other dependencies. You can do this with
    pip freeze > requirements.txt

Obviously, you should make sure you're in the virtualenv and you don't have any extra packages installed in the virtualenv, otherwise that'll be listed in the requirements file also.

tdowds:
Here's what it looks in its most updated form. Also I edited the link in AdvisorProfileView so that the invoice would open up in a new tab when you press the link.
screen shot 2015-09-22 at 8 12 44 am

nathanielparke:
I think this is okay I will be merging if there are no objections.

PR #419, opened 2015-09-24
Logmiddleware

tdowds:
What this does so far is it captures responses coming from requests sent to the api and logs the response and the request path to the database. There are some things we need to consider before moving forward:
1) What information do we want to log?
2) How do we want to present that information to the logger (kind of janky right now)?
3) Do we want to log to the console as well? (Doing this automatically when the unit tests we're running)

### DIFF
--- a/huxley/core/middlewares.py
+++ b/huxley/core/middlewares.py
@@ -15,5 +15,8 @@ class ServerLoggingMiddleware(object):
 
 class LoggingMiddleware(object):
     def process_response(self, request, response):
-        print(response)
+        if request.path.find('api') >= 0:
+            logger = logging.getLogger('huxley.db')
+            logger.info(request.path+', response data: '+str(response.data))
 
+        return response
--- a/huxley/core/middlewares.py
+++ b/huxley/core/middlewares.py
@@ -17,13 +17,10 @@ class LoggingMiddleware(object):
     def process_response(self, request, response):
         if 'api' in request.path:
             logger = logging.getLogger('huxley.api')
-            uri = request.path
-            message = "Logging response from the api."
-            status_code = response.status_code
             log = json.dumps({
-                  'message': message,
-                  'uri': uri,
-                  'status_code': status_code})
+                  'message': "Logging response from the api.",
+                  'uri': request.path,
+                  'status_code': response.status_code})
             logger.info(log)
 
         return response
--- a/huxley/core/middlewares.py
+++ b/huxley/core/middlewares.py
@@ -12,3 +12,8 @@ class ServerLoggingMiddleware(object):
     def process_exception(self, request, exception):
         logger = logging.getLogger('huxley.server')
         logger.exception(exception)
+
+class LoggingMiddleware(object):
+    def process_response(self, request, response):
+        print(response)
+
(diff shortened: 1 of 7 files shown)


huxley/core/middlewares.py
```diff
@@ -12,3 +12,11 @@ class ServerLoggingMiddleware(object):
     def process_exception(self, request, exception):
         logger = logging.getLogger('huxley.server')
         logger.exception(exception)
+
+class LoggingMiddleware(object):
+    def process_response(self, request, response):
+        if request.path.find('api') >= 0:
```
kmeht:
```python
if 'api' in request.path:
```

huxley/core/middlewares.py
```diff
@@ -12,3 +12,11 @@ class ServerLoggingMiddleware(object):
...
+            logger = logging.getLogger('huxley.db')
```
kmeht: This suggests it's a database logger, which doesn't make sense. Why not make it `huxley.api`?

kmeht:
I think we'd at least want to log the following for sure:
- the request URI
- the status code
- the API response data (assuming there's no sensitive information that we pass through there)
- the timestamp
- the username of the logged-in user

I'm rather unfamiliar with the logger, but if it just expects a string, we can format a string and pass it in. If there's a way to add more structured data (so we can store it in the DB), that'd allow us to actually query the DB with specific statements when we wanted to look up the logs for whatever reason (e.g. a particular user is having trouble, so we only want their logs).

kmeht:
@tdowds, so what's the plan with this?

tdowds:
Expanded the model to log the uri of the request and status code of the response object. Waiting to log response data and username of the logged in user just cause there's some edge cases I'm not sure how to handle yet (like what information in the response data do we want to log, and also how do we handle who the logged in user is when trying to log in) and also cause I want to make sure the approach I took with the uri and status code is the correct one

huxley/core/middlewares.py
```diff
@@ -12,3 +12,18 @@ class ServerLoggingMiddleware(object):
     def process_exception(self, request, exception):
         logger = logging.getLogger('huxley.server')
         logger.exception(exception)
+
+class LoggingMiddleware(object):
+    def process_response(self, request, response):
+        if 'api' in request.path:
+            logger = logging.getLogger('huxley.api')
+            uri = request.path
```
kmeht: Why is this assigned to a local variable? `uri`, `message` and `status_code` can be inlined into the dictionary.

huxley/logging/handlers.py
```diff
@@ -15,10 +16,16 @@ def emit(self, record):
 
         try:
             self.format(record)
+            data = json.loads(record.message)
+            message = data["message"]
```
kmeht: Same thing with inlining the local variables into the `LogEntry` call. Also, nit: use single quotes.

kmeht:
Looks good to me! @nathanielparke, this is good to merge after the small changes.

tdowds:
Updated

nathanielparke:
Closed via ffa5e6505ee683514df4330c53046d8a860a6d85

PR #424, opened 2015-11-14
logs the username of the request sent to the api.

tdowds:
Logs the username for each request sent to the api

### DIFF
--- a/huxley/core/middlewares.py
+++ b/huxley/core/middlewares.py
@@ -20,7 +20,8 @@ class LoggingMiddleware(object):
             log = json.dumps({
                   'message': "Logging response from the api.",
                   'uri': request.path,
-                  'status_code': response.status_code})
+                  'status_code': response.status_code,
+                  'username': request.user.username})
             logger.info(log)
 
         return response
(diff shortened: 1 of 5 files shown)


huxley/core/middlewares.py
```diff
@@ -20,7 +20,8 @@ def process_response(self, request, response):
             log = json.dumps({
                   'message': "Logging response from the api.",
                   'uri': request.path,
-                  'status_code': response.status_code})
+                  'status_code': response.status_code,
+                  'username': request.user.username})
```
kmeht: Will this throw if the user isn't logged-in? i.e. an anonymous user?

huxley/logging/models.py
```diff
@@ -11,6 +11,7 @@ class LogEntry(models.Model):
 
     uri = models.CharField(max_length=200)
     status_code = models.PositiveSmallIntegerField(default=0)
+    username = models.CharField(max_length=200)
```
kmeht: We need a migration for this too.

huxley/core/middlewares.py
tdowds: It returns an empty string

nathanielparke:
Merged via 72f76ae12f4b8836931f6ddbf3a3a0eb75114862

Issue #413, opened 2015-09-14
Generate Conference Dates and Session Dynamically


nathanielparke:
Currently our UI has hard coded dates and conference titles. To aid in quick deployment we should make any reference to a conference name, date or session dynamic. This also moves the project to a more modular application that can easily be ported between conferences rather than BMUN dependent.

nathanielparke:
closed via cac569d23b21e02ed68cffd5cff85e176cbc5a2e

PR #425, opened 2015-11-14
#413 dynamically generate confernce session on the front end.

tdowds:
This handles generating the session of the conference we're on. It was easier to send an integer up rather than the whole object (ran into some serialization issues with the conference model), but I wanted to start with this to first make any adjustments on the approach.

### DIFF
--- a/huxley/www/static/js/huxley/components/Huxley.js
+++ b/huxley/www/static/js/huxley/components/Huxley.js
@@ -34,7 +34,7 @@ var Huxley = React.createClass({
       return <RouteHandler user={user} />;
     } else if (User.isAdvisor(user)) {
       return (
-        <AdvisorView user={user}>
+        <AdvisorView user={user} conferenceSession={conferenceSession}>
           <RouteHandler user={user} />
         </AdvisorView>
       );
--- a/huxley/www/static/js/huxley/components/Huxley.js
+++ b/huxley/www/static/js/huxley/components/Huxley.js
@@ -17,6 +17,14 @@ var RouteHandler = Router.RouteHandler;
 var Huxley = React.createClass({
   mixins: [Router.Navigation],
 
+  childContextTypes: {
+    session: React.PropTypes.number
+  },
+
+  getChildContext: function() {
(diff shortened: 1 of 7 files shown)


huxley/www/views.py
```diff
@@ -21,6 +22,7 @@ def index(request):
 
     context = {
         'user_json': json.dumps(user_dict).replace('</', '<\\/'),
+        'conference_session': Conference.objects.filter(session=64)[0].session,
```
kmeht: Seems like an unnecessary fetch, since we're hardcoding the session to `64`. Can we just get the latest conference by session instead? That way it'll work for whatever conference we're at.

huxley/www/static/js/huxley/components/Huxley.js
```diff
@@ -34,7 +34,7 @@ var Huxley = React.createClass({
       return <RouteHandler user={user} />;
     } else if (User.isAdvisor(user)) {
       return (
-        <AdvisorView user={user}>
+        <AdvisorView user={user} conferenceSession={conferenceSession}>
```
kmeht: Where do we use this prop? Seems like we're using the global everywhere.

huxley/www/views.py
nathanielparke: I think that is ultimately the plan but this is just there for a filler while @tdowds builds the conferenceStore

kmeht:
If we're going to assume that conference will never change during a single session (which is reasonable to assume), we can use React's `context` feature to pass the session down through the entire component hierarchy. That way, we only have to refer to the global once, and all the components can access it via `this.context.conferenceSession`.

In general the feature should be using sparingly, but I believe immutable environment data is fair game.

https://facebook.github.io/react/docs/context.html

tdowds:
switched to using context

nathanielparke:
Looks good to me!

kmeht:
This is good to merge, whoever wants to do it. @nathanielparke?

tdowds:
closed by cac569d23b21e02ed68cffd5cff85e176cbc5a2e

PR #440, opened 2016-04-03
Country dropdowns

m-j-mcdonald:
Prevent selection of same country twice in registration

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 Austin Chen <akrolsmir@gmail.com>
 Evan Kawahara <Evan.Kawahara@disney.com>
 Kunal Mehta <k.mehta@berkeley.edu>
+Michael McDonald <m_j_mcdonald@comcast.net>
 Nathaniel Parke <nathaniel.parke@gmail.com>
 Pranay Patil <patilpranay@gmail.com>
 Trevor Dowds <tdowds@berkeley.edu>
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,4 +9,3 @@ Vin Ganeshan <ganeshan.vinayak@gmail.com>
 Vishal Kal <vishal.kal0@gmail.com>
 Vishal Kalyanasundaram <vishalkal@berkeley.edu>
 William Chieng <wchieng@berkeley.edu>
-m-j-mcdonald <m_j_mcdonald@comcast.net>
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,3 +8,4 @@ Vin Ganeshan <ganeshan.vinayak@gmail.com>
 Vishal Kal <vishal.kal0@gmail.com>
 Vishal Kalyanasundaram <vishalkal@berkeley.edu>
 William Chieng <wchieng@berkeley.edu>
+m-j-mcdonald <m_j_mcdonald@comcast.net>
--- a/huxley/www/static/js/huxley/components/CountrySelect.js
+++ b/huxley/www/static/js/huxley/components/CountrySelect.js
@@ -41,7 +41,7 @@ var CountrySelect = React.createClass({
   renderCommitteeOptions: function() {
     return this.props.countries.map(function(country) {
       if (!country.special) {
-        var index = this.props.countryPreferences.indexOf(""+country.id)
+        var index = this.props.countryPreferences.indexOf("" + country.id)
         return (
           <option key={country.id} value={country.id} disabled={index >= 0}>
             {country.name}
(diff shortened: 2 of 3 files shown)


AUTHORS
```diff
@@ -8,3 +8,4 @@ Vin Ganeshan <ganeshan.vinayak@gmail.com>
 Vishal Kal <vishal.kal0@gmail.com>
 Vishal Kalyanasundaram <vishalkal@berkeley.edu>
 William Chieng <wchieng@berkeley.edu>
+m-j-mcdonald <m_j_mcdonald@comcast.net>
```
nathanielparke: Change your git config so that your full name appears here instead of the handle

huxley/www/static/js/huxley/components/CountrySelect.js
```diff
@@ -3,6 +3,11 @@
  * Use of this source code is governed by a BSD License (see LICENSE).
  */
 
+/**
+ * Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+ * Use of this source code is governed by a BSD License (see LICENSE).
+ */
+
```
nathanielparke: We don't need a second header here

huxley/www/static/js/huxley/components/CountrySelect.js
```diff
@@ -34,14 +33,22 @@ var CountrySelect = React.createClass({
 
   renderCommitteeOptions: function() {
     return this.props.countries.map(function(country) {
-      if (!country.special) {
+      var index = this.props.countryPreferences.indexOf(""+country.id)
```
nathanielparke: preserve the feature that prevents seeing special countries

huxley/www/static/js/huxley/components/CountrySelect.js
```diff
@@ -34,14 +33,22 @@ var CountrySelect = React.createClass({
...
+      if(index < 0){
         return (
           <option key={country.id} value={country.id}>
             {country.name}
           </option>
         );
+
+      }else{
+        return (
```
kmeht: Why are we returning an entire other element if it's disabled? Does the following not work?
```
return (
  <option key={country.id} value={country.id} disabled={index >= 0}>
    {country.name}
  </option>
);
```

kmeht: And while we're here, small style nits, add spaces around keywords and binary operators:
```
var foo = bar.indexOf("" + country.id)
if (...) {
  ...
} else {
  ...
}
```
No need to update a diff just for style nits, but since there are substantive changes to be made, add these in there too :)

huxley/www/static/js/huxley/components/CountrySelect.js
```diff
@@ -11,14 +16,8 @@ var CountrySelect = React.createClass({
   propTypes: {
     onChange: React.PropTypes.func,
     countries: React.PropTypes.array,
-    selectedCountryID: React.PropTypes.number
-  },
-
-  shouldComponentUpdate: function(nextProps, nextState) {
```
kmeht: Why are we removing this method? It's an important optimization to prevent us from rendering 1920-ish (192 x 10) React elements with every single keystroke.

huxley/www/static/js/huxley/components/CountrySelect.js
```diff
@@ -11,13 +11,20 @@ var CountrySelect = React.createClass({
...
-    selectedCountryID: React.PropTypes.number
+    selectedCountryID: React.PropTypes.number,
+    countryPreferences: React.PropTypes.array
   },
 
   shouldComponentUpdate: function(nextProps, nextState) {
+    var preferenceChange = false;
+    for (var i = 0; i < this.props.countryPreferences.length; i++) {
+      if (this.props.countryPreferences[i] != nextProps.countryPreferences[i]) 
+        preferenceChange = true;
```
kmeht: If we know that the preferences have changed, we can return early, avoiding the `preferenceChange` variable, and falling back to that `return (... || ...)` statement otherwise.

```js
for (var i = 0; i < this.props.countryPreferences.length; i++) {
  if (this.props.countryPreferences[i] != nextProps.countryPreferences[i]) {
    return true;
  }
}
return (
  ... ||
);
```

Couple of style nits:
- prefer `===` and `!==` in JS, unless you specifically want type coercion. You can look up a famous talk on JS called "WAT" to see why.
- Always put braces after `if/for/while/etc` statements. Look up the Apple `goto fail` bug to see why.

tdowds:
closed via 0d80ab72e447ffbf9dd7ff9da06e31f02b4063e1

PR #445, opened 2016-04-10
Use standard flux in AssignmentStore.

tdowds:
Transitions to standard flux in the AssignmentStore. If this looks good, I'm gonna follow the same logic for the other remaining stores.

Only consideration is that since we must overwrite the callback that gets registered to the dispatcher in the store, do we want to use actions to call our functions like getAssignments? Not really sure what the benefit of doing that would be or if those benefits are worth the effort.

### DIFF
--- a/huxley/www/static/js/huxley/stores/__tests__/AssignmentStore-test.js
+++ b/huxley/www/static/js/huxley/stores/__tests__/AssignmentStore-test.js
@@ -16,12 +16,17 @@ describe('AssignmentStore', function() {
   beforeEach(function() {
     $ = require('jquery');
     AssignmentStore = require('../AssignmentStore');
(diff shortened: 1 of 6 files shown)


huxley/www/static/js/huxley/stores/__tests__/AssignmentStore-test.js
```diff
@@ -24,6 +25,10 @@ describe('AssignmentStore', function() {
     });
   });
 
+  it('subscribes to the dispatcher', function() {
+    expect(Dispatcher.register).toBeCalled();
+  });
+
```
tdowds: Not sure how necessary this is. Added it in because we're suppose to overwrite the callback that gets passed into `Dispatcher.register`.

tdowds:
Made changes for the other stores.

PR #455, opened 2016-05-17
Routing Bug On Assignment Finalization

nathanielparke:
This fixes the bug that was introduced with assignment finalization that rerouted the page to the advisor profile view, away from the assignments view. This was not desired behaviour.


nathanielparke:
10041cf7a8705c5bc1b5ee241c54decc55b10926

Issue #452, opened 2016-05-16
Admin panel exports csv with backward country preferences


tdowds:
On the admin panel on prod, whenever a csv is exported of every school's country preferences, their preferences are backwards. We need to investigate this and fix it.

tdowds:
Closed via d2a4aedb18528117b2ab077953a3a128564018a9

PR #460, opened 2016-05-20
#452 fixed backward country preferences issue

tdowds:
Enforces an ordering on the query to get country preferences when exporting those preferences to a csv


tdowds:
closed via d2a4aedb18528117b2ab077953a3a128564018a9

kmeht:
You might want to test this in staging to see if it actually fixed the behavior, since the bug only appears in production.

tdowds:
Working on that

PR #461, opened 2016-05-21
Fee Calculation

michaeleliot:
This calculates fees using number of assignments instead of number of delegates.

### DIFF
--- a/huxley/api/urls.py
+++ b/huxley/api/urls.py
@@ -27,6 +27,7 @@ urlpatterns += patterns('',
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
+    url(r'^schools/(?P<pk>[0-9]+)/assignments/finalize', views.school.SchoolAssignmentsFinalize.as_view(), name='school_assignments_finalize'),
     url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice')
 )
 
--- a/huxley/api/views/school.py
+++ b/huxley/api/views/school.py
@@ -45,8 +45,6 @@ class SchoolAssignments(generics.ListAPIView):
 
         return Assignment.objects.filter(school_id=school_id)
 
-<<<<<<< 5d3a71fa59dc25788b33464de26c80d1c46a7b39
-=======
 class SchoolAssignmentsFinalize(generics.GenericAPIView):
     authentication_classes = (SessionAuthentication,)
     permission_classes = (IsSchoolAdvisorOrSuperuser,)
@@ -66,7 +64,7 @@ class SchoolAssignmentsFinalize(generics.GenericAPIView):
 class SchoolInvoice(generics.CreateAPIView):
     authentication_classes = (SessionAuthentication,)
     permission_classes = (IsSchoolAdvisorOrSuperuser,)
->>>>>>> adds in api work for finalized assignments
+
 
 class SchoolInvoice(PDFTemplateView):
 
--- a/huxley/api/views/school.py
+++ b/huxley/api/views/school.py
@@ -45,6 +45,28 @@ class SchoolAssignments(generics.ListAPIView):
 
         return Assignment.objects.filter(school_id=school_id)
 
+<<<<<<< 5d3a71fa59dc25788b33464de26c80d1c46a7b39
+=======
+class SchoolAssignmentsFinalize(generics.GenericAPIView):
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsSchoolAdvisorOrSuperuser,)
+
+    def post(self, request, *args, **kwargs):
+        school_id = self.kwargs.get('pk', None)
+        school = School.objects.get(id=school_id)
+        school.assignments_finalized = True
+        school.save()
+        data = json.load(request)
+        for identity in data:
+            assignment = Assignment.objects.get(pk=int(identity))
+            assignment.delete()
+
+        return Response(status=status.HTTP_200_OK)
+
+class SchoolInvoice(generics.CreateAPIView):
+    authentication_classes = (SessionAuthentication,)
+    permission_classes = (IsSchoolAdvisorOrSuperuser,)
+>>>>>>> adds in api work for finalized assignments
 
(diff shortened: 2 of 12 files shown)


huxley/api/views/school.py
```diff
@@ -45,8 +45,6 @@ def get_queryset(self):
 
         return Assignment.objects.filter(school_id=school_id)
```
tdowds: Add a space here

huxley/api/views/school.py
```diff
@@ -66,7 +64,7 @@ def post(self, request, *args, **kwargs):
 class SchoolInvoice(generics.CreateAPIView):
```
tdowds: Is this class suppose to be here? It's not on the most updated version of master

huxley/core/models.py
```diff
@@ -150,11 +150,7 @@ def update_country_preferences(self, country_ids):
     @classmethod
     def update_fees(cls, **kwargs):
         school = kwargs['instance']
-        delegate_fees = cls.DELEGATE_FEE * sum((
-            school.beginner_delegates,
-            school.intermediate_delegates,
-            school.advanced_delegates,
-        ))
+        delegate_fees = cls.DELEGATE_FEE * (len(Assignment.objects.filter(school_id=school.id, is_double=False))+2*len(Assignment.objects.filter(school_id=school.id, is_double=True)))
```
tdowds: Since these fees are based off number of assignments, not number of delegates, we should rename `DELEGATE_FEE` to be `ASSIGNMENT_FEE`

huxley/core/tests/models/test_models.py
```diff
@@ -92,27 +92,44 @@ def test_update_country_preferences(self):
...
-        school.intermediate_delegates = i2
-        school.advanced_delegates = a2
+        Assignment.objects.bulk_create([
+            Assignment(committee_id=cm, country_id=ct, school_id=school.id, is_double=False)
+            for ct in [ct1, ct2]
+            for cm in [cm1, cm2]
+        ])
+
+        #test if fee is calculated using assingments 
         school.save()
```
tdowds: By saving the school after we add/update assignments this implies that every time we add/update assignments we must save the school in order to recalculate its fees. It might make more sense to recalculate the fees automatically any time an assignment is added or deleted. Some ways of doing this is signaling after an Assignment object is created or deleted to recalculate a school's fees. Or in the update_assignment functions modify it to call the update_fees method in the School class directly.

huxley/core/tests/models/test_models.py
```diff
@@ -92,27 +92,44 @@ def test_update_country_preferences(self):
...
+        self.assertEquals(
+            school.fees_owed,
+            school.REGISTRATION_FEE + school.DELEGATE_FEE * 4,
```
tdowds: Will need to modify this to be `ASSIGNMENT_FEE`

huxley/core/tests/models/test_models.py
```diff
@@ -92,27 +92,44 @@ def test_update_country_preferences(self):
...
+        )
 
+        Assignment.objects.bulk_create([
+            Assignment(committee_id=cm, country_id=ct, school_id=school.id, is_double=True)
+            for ct in [ct3]
+            for cm in [cm1, cm2]
+        ])
+
+        #tests if it detects difference between single and double delegates (double should pay twice as much)
+        school.save()
```
tdowds: Shouldn't need this line if we automatically update fees once assignments are added or deleted.

huxley/core/tests/models/test_models.py
```diff
@@ -92,27 +92,44 @@ def test_update_country_preferences(self):
...
         self.assertEquals(
             school.fees_owed,
-            school.REGISTRATION_FEE + school.DELEGATE_FEE * (b2 + i2 + a2),
+            school.REGISTRATION_FEE + school.DELEGATE_FEE * 8,
```
tdowds: Will need to adjust this to be `ASSIGNMENT_FEE`

tdowds:
Good first diff! Not sure where the edit in `huxley/api/views/schools.py` came from but you probably don't need that. Consider these comments, but there is more work to be done on alerting the user that fees are calculated once assignments are given out.

huxley/api/urls.py
```diff
@@ -27,6 +27,7 @@
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
+    url(r'^schools/(?P<pk>[0-9]+)/assignments/finalize', views.school.SchoolAssignmentsFinalize.as_view(), name='school_assignments_finalize'),
```
nathanielparke: I think this indicates that you might have an outdated branch. Make sure that 'fab update' runs in the 'fab submit' script so you get the latest version of the branch.

huxley/api/views/school.py
```diff
@@ -45,6 +45,26 @@ def get_queryset(self):
...
+        school_id = self.kwargs.get('pk', None)
+        school = School.objects.get(id=school_id)
+        school.assignments_finalized = True
+        school.save()
+        data = json.load(request)
+        for identity in data:
+            assignment = Assignment.objects.get(pk=int(identity))
+            assignment.delete()
+
+        return Response(status=status.HTTP_200_OK)
```
nathanielparke: Same as above. Branch coherency is key.

huxley/core/admin/assignment.py
```diff
@@ -59,7 +59,8 @@ def generate_assigments(reader):
                     rejected = False # allow for the rejected field to be null
                 else:
                     rejected = (row[3].lower() == 'true') # use the provided value if admin provides it
-                yield (committee.id, country.id, school.id, rejected)
+                is_double = True
```
nathanielparke: This is wrong. What this will do is set every single assignment to have the 'is_double' field as True. This isn't the behavior we want because not every assignment is a double delegation. What probably needs to happen is we need to change our CSV schema for uploading assignments to have a is_double field. We can then set the is_double field to whatever value is passed in by external.

nathanielparke:
Couple of things. First squash all those migrations into one migration using 'python manage.py squashmigrations'. Second, I'm not a huge fan of naming the field 'is_double'. I think we might consider either changing the name to something like 'dual' or 'double'. 'is_double' seems clunky to me. Another thing to possibly think about is if we want this field to be a numerical field so that it can handle larger than double delegations. This is probably not something that we at BMUN will use but it might be relevant to other conferences that use our code.

tdowds:
Closed due to change in preferences from treasurer and SG

Issue #457, opened 2016-05-18
Add Conference Dependent Data to Conference Model


nathanielparke:
We should add in things like current Secretary General, Current External, the price per delegate for the current year etc. to the conference model so it can all be tracked through the admin panel and none of that is hard coded. This will require propagating the data through to things such as the AdvisorProfileView using Flux

tdowds:
Todos for this:
- [ ] Propagate start and end date to LoginView
- [ ] Propagate name of external to AdvisorProfileView and RegistrationSuccessView
- [ ] Propagate fees to RegistrationSuccessView
- [ ] Use fees in School model
- [ ] Use session in email confirmation

tdowds:
Closed via 1507f8e836677805874fcf11bf3ffa5eabd709d9

PR #466, opened 2016-05-30
#457 included start and end date and passing conference specific details as object to client.

tdowds:


huxley/www/views.py
```diff
@@ -21,9 +21,17 @@ def index(request):
     if request.user.is_authenticated():
         user_dict = UserSerializer(request.user).data
 
+    session = Conference.objects.all().aggregate(Max('session'))['session__max']
+    conference = Conference.objects.get(session=session)
```
nathanielparke: I feel like this implementation is bug prone. If we don't explicitly document that we are using the maximum conference then users could see some unexpected behavior. I am okay leaving it as is but we would need to make it clear. One thing I have been thinking about is to create a properties file that stores all these properties about the app then injects them into the backend. That will take some more thought but this would be a place where that is useful.

huxley/www/static/js/huxley/components/Huxley.js
```diff
@@ -18,11 +18,24 @@ var Huxley = React.createClass({
...
+    external: React.PropTypes.string,
+    registrationFee: React.PropTypes.number,
+    delegateFee: React.PropTypes.number
   },
 
   getChildContext: function() {
-    return {session: global.conferenceSession}
+    var conference = global.conference;
+    return {
+      session: conference['session'],
```
kmeht: perhaps we should just pass the conference down as context? we'd have to be careful not to mutate it anywhere, but it beats passing down every single field individually.

huxley/core/models.py
```diff
@@ -15,16 +15,23 @@
...
+    min_attendance   = models.PositiveSmallIntegerField(default=0)
+    max_attendance   = models.PositiveSmallIntegerField(default=0)
+    open_reg         = models.BooleanField(default=True)
+    waitlist_reg     = models.BooleanField(default=False)
+    external         = models.CharField(max_length=128)
+    registration_fee = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('50.00'))
+    delegate_fee     = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('50.00'))
+
+    @classmethod
+    def get_conference(cls):
```
kmeht: nit: I think `get_current` is a more appropriate name (it's not clear what conference you're getting with `get_conference`).

huxley/api/views/school.py
```diff
@@ -68,9 +69,9 @@ def get(self, request, *args, **kwargs):
             "date_registered": school.registered.strftime("%m/%d/%y"),
             "due_date": due_date.strftime("%m/%d/%y"),
             "delegate_total": delegate_total,
-            "delegate_fee": School.DELEGATE_FEE,
+            "delegate_fee": delegate_fee,
             "delegate_fees": delegate_fees,
-            "registration_fee": School.REGISTRATION_FEE,
+            "registration_fee": Conference.get_conference().registration_fee,
```
kmeht: We should assign to a local variable if we're gonna be using the conference multiple times :)

huxley/core/tests/models/test_models.py
```diff
@@ -99,9 +101,11 @@ def test_update_fees(self):
             advanced_delegates=a,
         )
 
+        registration_fee = Conference.get_conference().registration_fee
+        delegate_fee = Conference.get_conference().delegate_fee
```
kmeht: local variable. (I don't mean to be nitty, but I don't want to rely on the ORM's caching behavior to that degree).

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -25,7 +25,7 @@ var AdvisorAssignmentsView = React.createClass({
   ],
 
   contextTypes: {
-    session: React.PropTypes.number
+    conference: React.PropTypes.object
```
kmeht: check out `PropTypes.shape`. It allows you to specify the types of each property of the object: https://facebook.github.io/react/docs/reusable-components.html#prop-validation

You can define the proptypes in another module (e.g. ConferenceContext) and `require` them so you don't have to write out the entire shape everywhere you have to write `contextTypes` for a component.

huxley/www/static/js/huxley/components/Huxley.js
```diff
@@ -18,11 +18,14 @@ var Huxley = React.createClass({
...
-    session: React.PropTypes.number
+    conference: React.PropTypes.object
   },
 
   getChildContext: function() {
-    return {session: global.conferenceSession}
+    var conference = global.conference;
+    return {
+      conference: conference
+    }
```
kmeht: semicolon

huxley/core/migrations/0011_auto_20160602_2136.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
nathanielparke: You can probably squash these two auto migrations together

nathanielparke:
This is awesome!

tdowds:
closed via 1507f8e836677805874fcf11bf3ffa5eabd709d9

PR #468, opened 2016-05-30
Advisor roster view additions

jmosky12:
*Added navbar link
*Created Add Delegate Button and Delete Delegate Button


huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -31,7 +31,7 @@ var AdvisorRosterView = React.createClass({
       delegates: [],
       committees: {},
       countries: {},
-      loading: false
+      loading: false,
```
tdowds: No comma

jmosky12:
ui

tdowds:
@jmosky12 add a footer to the table and put the add delegate button in there. Take a look at the code in `AdvisorAssignmentsView.js` as a reference.

nathanielparke:
@jmosky12, I think it might make sense to have each table row drop down when clicked to reveal more information. second example might be relevant: http://stackoverflow.com/questions/16926752/expand-collapse-table-rows-with-jquery, or this one is also nice: http://jsfiddle.net/mfwYS/

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -32,6 +32,10 @@ var AdvisorRosterView = React.createClass({
       committees: {},
       countries: {},
       loading: false,
+      errors: {},
+      name: '',
+      email: '',
+      summary: '',
```
tdowds: No comma

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -68,7 +72,8 @@ var AdvisorRosterView = React.createClass({
       <InnerView>
         <h2>Roster</h2>
         <p>
-          Here you can add your school's delegates to your roster.
+          REVERT BACK
+          // Here you can add your schools delegates to your roster.
```
tdowds: Uncomment this and remove the line above

huxley/api/urls.py
```diff
@@ -21,7 +21,9 @@
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
-    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice'),
+    url(r'^schools/(?P<pk>[0-9]+)/delegates/?$', views.school.SchoolDelegates.as_view(), name='school_delegates'),
+    url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice')
+)
```
tdowds: This parentheses needs to be removed and a comma needs to be placed at the end of the line above it

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -109,6 +114,51 @@ var AdvisorRosterView = React.createClass({
     }.bind(this));
   },
 
+  _addDelegatePressed: function() {
+    return (
+      <form>
+        Name: <input type="text" placeholder="Name" valueLink={this.linkState('name')} /><br>
```
tdowds: `br` is a self closing element. Should be `<br />` . That should fix your compiler error you were mentioning

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -109,6 +114,51 @@ var AdvisorRosterView = React.createClass({
...
+
+  _handleSubmit: function(data) {
+    this.setState({loading: true});
+    $.ajax({
+      type: 'POST',
+      url: '/api/delegate',
+      data: {
+        name: this.state.name,
+        email: this.state.email,
+        summary: this.state.summary
```
tdowds: Advisors shouldn't be able to submit a summary. The summary should be read only.

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -109,6 +114,51 @@ var AdvisorRosterView = React.createClass({
     }.bind(this));
   },
 
+  _addDelegatePressed: function() {
+    return (
+      <form>
+        Name: <input type="text" placeholder="Name" valueLink={this.linkState('name')} /><br>
+        Email: <input type="text" placeholder="Email" valueLink={this.linkState('email')}/><br>
+        Summary: <input type="text" placeholder="Summary" valueLink={this.linkState('summary')}/><br>
```
tdowds: No need for inputting a summary. This is a feature advisors don't have control over. Remove this line

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -117,9 +117,9 @@ var AdvisorRosterView = React.createClass({
   _addDelegatePressed: function() {
     return (
       <form>
-        Name: <input type="text" placeholder="Name" valueLink={this.linkState('name')} /><br>
-        Email: <input type="text" placeholder="Email" valueLink={this.linkState('email')}/><br>
-        Summary: <input type="text" placeholder="Summary" valueLink={this.linkState('summary')}/><br>
+        <br>Name: <input type="text" placeholder="Name" valueLink={this.linkState('name')} /></br>
+        <br>Email: <input type="text" placeholder="Email" valueLink={this.linkState('email')}/></br>
+        <br>Summary: <input type="text" placeholder="Summary" valueLink={this.linkState('summary')}/></br>
```
tdowds: This line needs to be removed

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -0,0 +1,90 @@
...
+from huxley.core.models import Assignment, Delegate, School
+from huxley.utils.test import (TestCommittees, TestCountries, TestSchools,
+                               TestUsers, TestAssignments, TestDelegates)
+
+
+class SchoolDelegateGetTestCase(ListAPITestCase):
+    url_name = 'api:school_delegates'
+    is_resource = True
+
+    def setUp(self):
```
kmeht: It looks like we're not modifying these objects in between test cases. In that case, we can use the class method, which is run only once (instead of before every test). This should make the test run a bit faster: 

https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.TestCase.setUpTestData

Furthermore, we don't have to make all of these objects class/instance members. For example, we only use `committee1` in this method, so it can just be a local variable instead of a `self.committee1` property.

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def assert_delegate_equal(self, response):
+        '''Assert that the response contains the assignments in order.'''
+        response.data[0].pop('created_at')
+        response.data[1].pop('created_at')
+        self.assertEqual(dict(response.data[0]),
+            {
+                'id': self.delegate1.id,
+                'assignment':self.assignment1.id,
+                'name':unicode(self.delegate1.name),
```
kmeht: nit: space after colon, same for all properties in this dictionary

huxley/api/urls.py
```diff
@@ -21,8 +21,8 @@
     url(r'^schools/?$', views.school.SchoolList.as_view(), name='school_list'),
     url(r'^schools/(?P<pk>[0-9]+)/?$', views.school.SchoolDetail.as_view(), name='school_detail'),
     url(r'^schools/(?P<pk>[0-9]+)/assignments/?$', views.school.SchoolAssignments.as_view(), name='school_assignments'),
+    url(r'^schools/(?P<pk>[0-9]+)/delegates/?$', views.school.SchoolDelegates.as_view(), name='school_delegates'),
     url(r'^schools/(?P<pk>[0-9]+)/invoice/?$', views.school.SchoolInvoice.as_view(), name='school_invoice'),
-
```
kmeht: Maintain this empty line, it separates the URIs for each model.

huxley/settings/conference.py
```diff
@@ -3,4 +3,4 @@
 
 CONFERENCE_WAITLIST_OPEN = False
 
-SESSION = 65
+SESSION = 69
```
kmeht: revert

huxley/utils/test.py
```diff
@@ -105,7 +105,11 @@ def new_country(**kwargs):
 class TestDelegates():
     @staticmethod
     def new_delegate(**kwargs):
-        a = TestAssignments.new_assignment()
+        if 'assignment' in kwargs.keys():
```
kmeht: A more idiomatic way of writing this is:
```python
a = kwargs.pop('assignment') or TestAssignments.new_assignment()
```

kmeht: Also, the `in` operator already operates on keys for dictionaries, so if you need to check if a key's in a dict, all you need is `'key' in my_dict`.

huxley/utils/test.py
```diff
@@ -117,9 +121,20 @@ def new_delegate(**kwargs):
 class TestAssignments():
     @staticmethod
     def new_assignment(**kwargs):
-        test_committee = TestCommittees.new_committee()
-        test_school = TestSchools.new_school()
-        test_country = TestCountries.new_country()
+        if 'committee' in kwargs.keys():
```
kmeht: Same as above for all three of these

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -0,0 +1,106 @@
...
+  componentWillMount: function() {
+    var user = CurrentUserStore.getCurrentUser();
+    AssignmentStore.getAssignments(user.school.id, function(assignments) {
+      this.setState({assignments: assignments.filter(
+        function(assignment) {
+          return !assignment.rejected
+        }
+      )});
+    }.bind(this));
+    CommitteeStore.getCommittees(function(committees) {
```
tdowds: Is there a reason to make an ajax call to get our committees when this page is loaded?

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -0,0 +1,106 @@
...
+      var new_committees = {};
+      for (var i = 0; i < committees.length; i++) {
+        new_committees[committees[i].id] = committees[i];
+      }
+      this.setState({committees: new_committees});
+    }.bind(this));
+    CountryStore.getCountries(function(countries) {
```
tdowds: Same as above

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -74,6 +74,7 @@ def assert_delegate_equal(self, response):
             {
                 'id': self.delegate1.id,
                 'assignment':self.assignment1.id,
+                'school':self.delegate1.school.id,
```
tdowds: Space here?

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -56,6 +56,7 @@ def test_other_user(self):
         self.client.login(username='another', password='user')
 
         response = self.get_response(self.school.id)
+        print response
```
tdowds: Probably should remove these print statements

kmeht:
What's the status on this PR, still active? Certainly needs to be rebased in light of recent changes.

huxley/www/static/js/huxley/stores/DelegateStore.js
```diff
@@ -0,0 +1,40 @@
...
+
+var $ = require('jquery');
+var Promise = require('es6-promise').Promise;
+var Dispatcher = require('../dispatcher/Dispatcher');
+var {Store} = require('flux/utils');
+
+
+var _delegatePromise = null;
+
+class DelegateStore extends Store {
```
tdowds: As a reminder, we should probably include a Jest test for this store

nathanielparke:
This PR is still in the works. Finding time for two people to work isn't an easy task.

tdowds:
Totally get that. Mostly just putting it there so we don't forget. Also if there are pieces you guys think you can delegate out let me know. I can add some more manpower to this task if that's what you need. Maybe I can work on that test? Would that be helpful?

kmeht:
@nathanielparke, @jmosky12: Couple of general comments on JS as you guys continue working on this. Two ES2015 features in particular will make the code much easier to read.

```js
/* Arrow functions (which automatically bind "this") */
// Before
foo.map(function(item) {
  return this.doSomething(item);
}.bind(this));

// After
foo.map((item) => this.doSomething(item));
// Or...
foo.map((item) => {
  return this.doSomething();
});

/* Object shorthand properties */
// Before
this.setState({foo: foo});
// After
this.setState({foo});
```

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -0,0 +1,142 @@
...
+    return {
+      assignments: [],
+      delegates: [],
+    };
+  },
+
+  componentWillMount: function() {
+    var user = CurrentUserStore.getCurrentUser();
+    AssignmentStore.getAssignments(user.school.id, function(assignments) {
+      this.setState({assignments: assignments.filter(
```
kmeht: Example of my last comment:
```js
this.setState({
  assignments: assignments.filter((a) => !a.rejected),
});
```

huxley/core/models.py
```diff
@@ -356,11 +356,11 @@ class Meta:
 
 class Delegate(models.Model):
     school        = models.ForeignKey(School, related_name='delegates', null=True)
-    assignment    = models.ForeignKey(Assignment, related_name='delegates', blank=True)
+    assignment    = models.ForeignKey(Assignment, related_name='delegates', blank=True, null=True)
     name          = models.CharField(max_length=64)
-    email         = models.EmailField(blank=True)
+    email         = models.EmailField(blank=True, null=True)
```
tdowds: Is there a reason why you have `null=True` here? I don't think it's necessary for this field to not be required.

huxley/core/models.py
```diff
@@ -356,11 +356,11 @@ class Meta:
...
     created_at    = models.DateTimeField(auto_now_add=True)
-    summary       = models.TextField(default='', null=True)
+    summary       = models.TextField(default='', blank=True, null=True)
```
tdowds: We should remove `null=True`. Unless we have some preference for this being null instead of an empty string when nothing is submitted.

https://docs.djangoproject.com/en/1.8/ref/models/fields/#null

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -89,22 +127,61 @@ var AdvisorRosterView = React.createClass({
...
-  // _handleAddDelegate: function() {
-  //   return (
-  //     <form>
-  //       <br>Name: <input type="text" placeholder="Name" valueLink={this.linkState('name')} /></br>
-  //       <br>Email: <input type="text" placeholder="Email" valueLink={this.linkState('email')}/></br>
-  //       <input type="submit" value="Submit" onclick={this._handleSubmit} />
-  //     </form>
-  //   )
-  // },
+  _handleAddDelegate: function() {
```
tdowds: Is this being called anywhere else?

huxley/www/static/js/huxley/components/AdvisorRosterView.js
```diff
@@ -89,22 +127,61 @@ var AdvisorRosterView = React.createClass({
...
+  },
+
+  _handleDelegateDeleteSuccess: function(data, status, jqXHR) {
+    var delegates = this.state.delegates
+    delegates = delegates.filter(function (delegate) {
+      return delegate.id != jqXHR.responseJSON.delegate.id
+    });
+    this.setState({delegates: delegates})
+    this.setState({loading: false});
+    this.history.pushState(null, '/advisor/roster');
```
tdowds: Will this cause the page to refresh and reload the delegates, or do we have to explicitly filter the delegates above like we do?

tdowds:
Is there a screenshot of our fancy new modal that you can share with us?

huxley/core/models.py
nathanielparke: This needs to have null=True for the admin panel. We should be able to add a delegate with a null email and we currently cannot do that because of admin panel validation.

huxley/core/models.py
nathanielparke: same as below. We want to be able to add delegates properly through the admin panel.

tdowds:
For the changes on the AdvisorAssignmentView these are the requirements I set for it.
1. Absolutely no delegate selection should be available until assignments are finalized.
2. After assignments are finalized, advisors can then select which delegates they want to assign to which assignment.
3. Each dropdown should be populated with all the delegates associated with that particular school
4. When an advisor selects a delegate that delegate cannot be selected in any other dropdown and is assigned to that assignment
5. If the selected delegate is replacing a delegate that was already in that particular dropdown that delegate should no longer be assigned to that assignment and can be picked in the other dropdowns
6. If none is selected in a dropdown, whatever delegate was assigned to that assignment is no longer assigned to that assignment and can be selected in other dropdowns
7. When the page is loaded initially, any delegate with an assignment should be selected in one of the assignment's dropdowns and that delegate cannot be selected in any other dropdown

tdowds:
To perform bulk updates I wrote a custom mixin that basically takes a list of delegates and updates them one at a time. There exists a package out there that does this that was super helpful. Also django rest framework's source code was really helpful as well. Here are the links to both that I used:

[UpdateModelMixin](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/mixins.py#L61)
[get_object](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L76)
[BulkUpdateModelMixin](https://github.com/miki725/django-rest-framework-bulk/blob/master/rest_framework_bulk/drf2/mixins.py#L45)

The way I implemented this is kind of hacky so any feedback would be much appreciated.

Some things that are still left to do.

1. Test the mixin and the overrided save functionality
2. Add some indication on the ui of a successful change
3. Add some indication on the ui of an error

I'll push up 1 on the next diff. 2 and 3 I think we should save for a different PR, and focus instead on merging this PR in.

tdowds:
DRF doesn't support responses from delete requests that contain any content
http://www.django-rest-framework.org/api-guide/generic-views/#destroymodelmixin

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -0,0 +1,245 @@
...
+    def assert_delegate_equal(self, response):
+        '''Assert that the response contains the delegates in order.'''
+        response.data[0].pop('created_at')
+        response.data[1].pop('created_at')
+        self.assertEqual(dict(response.data[0]),
+            {
+                'id': self.delegate1.id,
+                'assignment':self.assignment1.id,
+                'school':self.delegate1.school.id,
+                'name':unicode(self.delegate1.name),
```
kmeht: Does the test fail if these `unicode()` calls aren't there? If it still passes, let's remove them.

huxley/core/models.py
```diff
@@ -372,10 +374,12 @@ def country(self):
     def committee(self):
         return self.assignment.committee
 
-    @property
-    def school(self):
-        return self.assignment.school
+    def save(self, *args, **kwargs):
+        if (self.assignment and self.school and self.school.id != self.assignment.school.id):
```
kmeht: Models also have `_id` fields for their foreign keys so that you can compare keys without having to load the objects themselves. So you can write this as follows:
```python
if self.assignment_id and self.school_id and self.school_id != self.assignment.school_id:
    raise ValidationError(...)
```
This saves us from loading at max two school objects.

huxley/api/mixins.py
```diff
@@ -0,0 +1,43 @@
...
+        # few lines are trying to do. Any suggestions would be appreciated.
+        data = request.data
+        # This if statement is here for testing purposes
+        if isinstance(data, QueryDict):
+            data = json.loads(request.data.items()[0][0])
+
+        self.lookup_url_kwarg = 'obj_id'
+
+        response_data = []
+        for obj in data:
```
kmeht: We should batch this code, otherwise we're doing two round trips to DB for every single updated object: fetching it, then saving it.

Instead, we should fetch all the objects in one round with the `in` modifier on QuerySet: https://docs.djangoproject.com/en/1.9/ref/models/querysets/#in

Then, we iterate over them and update them inside a transaction, so all the saves occur in one DB request. We do this with the `transaction.atomic` context manager: https://docs.djangoproject.com/en/1.8/topics/db/transactions/#django.db.transaction.atomic An example of this is in the assignment updating code.

huxley/api/mixins.py
```diff
@@ -0,0 +1,43 @@
...
+            self.kwargs[self.lookup_url_kwarg] = obj['id']
+            instance = self.get_object()
+            serializer = self.get_serializer(instance, data=obj, partial=partial)
+            serializer.is_valid(raise_exception=True)
+            self.perform_list_update(serializer)
+            response_data.append(serializer.data)
+
+        return Response(response_data, status=status.HTTP_200_OK)
+
+    def perform_list_update(self, serializer):
```
kmeht: Does this need to be a separate function? Is it an override?

huxley/api/mixins.py
```diff
@@ -0,0 +1,43 @@
...
+        serializer.save()
+
+    def partial_list_update(self, request, *args, **kwargs):
```
kmeht: I don't think we need this method, we could just call `list_update` with the `partial` argument set from inside the view:
```python
def patch(self, request, *args, **kwargs):
    self.list_update(request, partial=True, *args, **kwargs)
```

huxley/api/tests/school/test_school_delegates.py
tdowds: They fail without the `unicode()` calls.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -56,11 +60,33 @@ var AdvisorAssignmentsView = React.createClass({
...
+      // assigned is to manage the relation between an assignment and its delegates.
+      var assigned = {};
+      for (var i = 0; i < delegates.length; i++) {
+        if (delegates[i].assignment) {
+          if (delegates[i].assignment in assigned) {
+            assigned[delegates[i].assignment][1] = delegates[i].id;
+          } else {
+            // We assume each assignment has at most two delegates assigned to it.
+            // Single delegate assignments simply won't access the second slot.
+            var slots = new Array(2).fill(0);
```
kmeht: Keep it simple:
```js
var slots = [0, 0];
```

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -56,11 +60,33 @@ var AdvisorAssignmentsView = React.createClass({
...
       var new_countries = {};
-      for (var i = 0; i <countries.length; i++) {
+      for (var i = 0; i < countries.length; i++) {
         new_countries[countries[i].id] = countries[i];
       }
       this.setState({countries: new_countries})
     }.bind(this));
+    DelegateStore.getDelegates(user.school.id, function(delegates) {
+      // assigned is to manage the relation between an assignment and its delegates.
+      var assigned = {};
```
kmeht: Might be worth extracting this into a function with an explanatory docblock. At first glance, I have no idea what this is attempting to do.

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -56,11 +60,33 @@ var AdvisorAssignmentsView = React.createClass({
     }.bind(this));
     CountryStore.getCountries(function(countries) {
       var new_countries = {};
-      for (var i = 0; i <countries.length; i++) {
+      for (var i = 0; i < countries.length; i++) {
         new_countries[countries[i].id] = countries[i];
       }
       this.setState({countries: new_countries})
     }.bind(this));
+    DelegateStore.getDelegates(user.school.id, function(delegates) {
```
kmeht: To avoid explicit bind, we can use arrow functions from ES2015:
```js
DelegateStore.getDelegates(user.school.id, (delegates) => {
  // do stuff...
}); // <- no more .bind(this)
```

package.json
```diff
@@ -15,6 +15,7 @@
     "js-cookie": "~2.0.3",
     "react": "15.0.0",
     "react-dom": "15.0.0",
+    "react-modal": "1.4.0",
```
kmeht: We only need one of these, right?

huxley/api/mixins.py
```diff
@@ -0,0 +1,45 @@
...
+        response_data = []
+        delegate_ids = []
+        updates = {}
+
+        for obj in data:
+            delegate_ids.append(obj['id'])
+            updates[obj["id"]] = obj
+
+        with transaction.atomic():
+            delegates = Delegate.objects.filter(id__in=updates)
```
kmeht: I imagine this is supposed to be `delegate_ids`, because it's not used anywhere in this class.

huxley/api/mixins.py
```diff
@@ -0,0 +1,45 @@
...
+        partial = kwargs.pop('partial', False)
+        data = request.data
+        if isinstance(data, QueryDict):
+            data = json.loads(request.data.items()[0][0])
+
+        response_data = []
+        delegate_ids = []
+        updates = {}
+
+        for obj in data:
```
kmeht: Let's go simpler with this. Assuming the data is an array of delegates, we can convert it to a dictionary with a dictionary comprehension:
```python
updates = {delegate['id']: delegate for delegate in data}
```
Then, in the transaction, we can use the keys of the dictionary:
```python
delegates = Delegate.objects.filter(id__in=updates.keys())
```

huxley/www/static/js/huxley/stores/DelegateStore.js
```diff
@@ -0,0 +1,40 @@
...
+
+'use strict';
+
+var $ = require('jquery');
+var Promise = require('es6-promise').Promise;
+var Dispatcher = require('../dispatcher/Dispatcher');
+var {Store} = require('flux/utils');
+
+
+var _delegatePromise = null;
```
kmeht: Since we're passing in a school id, we should actually have an object of promises per ID:
```js
var _delegatePromises = {};
...
if (!_delegatePromises[schoolID]) {
  _delegatePromises[schoolID] = new Promise(...);
}
```
Otherwise, no matter what schoolID you pass in, you'll get the delegates from the first school!

huxley/www/static/js/huxley/components/DelegateSelect.js
```diff
@@ -0,0 +1,39 @@
...
+
+'use strict';
+
+var React = require('react');
+
+var DelegateSelect = React.createClass({
+  propTypes: {
+    onChange: React.PropTypes.func,
+    delegates: React.PropTypes.array,
+    selectedDelegateId: React.PropTypes.number
```
kmeht: nit: `selectedDelegateID`

huxley/www/static/js/huxley/components/DelegateSelect.js
```diff
@@ -0,0 +1,39 @@
...
+        onChange={this.props.onChange}
+        value={this.props.selectedDelegateId}>
+        <option value="0">None</option>
+        {this.renderDelegateOptions()}
+      </select>
+    );
+  },
+
+  renderDelegateOptions: function() {
+    return this.props.delegates.map(function(delegate) {
```
kmeht: Little easier to read with an arrow function:
```js
return this.props.delegates.map((delegate) =>
  <option ...>
    {delegate.name}
  </option>
);
```
Also, this function doesn't even use `this`, so we didn't need to bind in the first place ;)

huxley/www/static/js/huxley/components/AdvisorView.js
```diff
@@ -37,6 +37,9 @@ var AdvisorView = React.createClass ({
           <NavTab href="/advisor/assignments">
             Assignments
           </NavTab>
+          <NavTab href="/advisor/roster">
+            Roster
```
kmeht: "Delegates" might be a clearer name to advisors.

huxley/api/mixins.py
```diff
@@ -0,0 +1,45 @@
...
+from rest_framework import status
+from rest_framework.response import Response
+
+class ListUpdateModelMixin(object):
+    """
+    Update a queryset
+    """
+
+    def list_update(self, request, *args, **kwargs):
+        partial = kwargs.pop('partial', False)
```
kmeht: Pretty sure this can just be an argument to the function:
```python
def list_update(self, request, partial=False, *args, **kwargs):
    ...
```

PR #469, opened 2016-06-10
432 improve assignment csv upload

m-j-mcdonald:
If there's an issue with an assignment upload this will create an error message with a list of failed rows  and abort the upload. This modifies the AssignmentAdmin to check if a row has bad entries and pass bad entries to the Assignment Model in a different format, then has the update_assignments method in the Assignment Model mark those entries, as well as check for duplicate assignments in the csv. If the update_assignment method receives any bad assignments it will not upload any assignments and instead will return a list of the failed assignments. The AssignmentAdmin will then take this list if it gets created, and create an error message out of the failed assignments.


huxley/core/admin/assignment.py
```diff
@@ -39,30 +41,69 @@ def load(self, request):
...
+                    row.append("") # extends the row to have the minimum proper num of columns
+
+                bad_columns = get_bad_columns(row)
+                already_assigned = (row[1], row[2]) in assigned
+                
                 if len(row) < 4:
                     rejected = False # allow for the rejected field to be null
                 else:
                     rejected = (row[3].lower() == 'true') # use the provided value if admin provides it
-                yield (committee.id, country.id, school.id, rejected)
```
tdowds: We should most definitely return a generator from this function. It takes up a lot of space otherwise

huxley/core/admin/assignment.py
```diff
@@ -39,30 +41,69 @@ def load(self, request):
...
+                    failed_rows.append(str((row[0], row[1], row[2], str(rejected))))
+
+        create_assignments(reader)
+        if not failed_rows:
+            Assignment.update_assignments(successful_rows)
+            return HttpResponseRedirect(reverse('admin:core_assignment_changelist'))
+        else:
+            messages.error(request, 
+                'Assignments aborted.\nThese are the rows that failed with identification of the cells at fault:\n' + ', \n'.join(failed_rows))
+            return HttpResponseRedirect(reverse('admin:core_assignment_changelist'))
```
tdowds: You can take this return outside the scope of this `if else` statement

huxley/core/admin/assignment.py
```diff
@@ -39,30 +41,69 @@ def load(self, request):
         '''Loads new Assignments.'''
         assignments = request.FILES
         reader = csv.reader(assignments['csv'])
+        successful_rows = []
```
tdowds: So instead of keeping track of successful rows I would try keeping the generator that is returned from `generate_assignments` and instead put a try catch in there so that every time an error occurs we catch it, append the row to `failed_rows` and then continue on in our iteration

tdowds:
Can you post a screenshot of what the output looked like?

m-j-mcdonald:
Screenshot:
screen shot 2016-06-11 at 8 26 26 pm

huxley/core/admin/assignment.py
```diff
@@ -41,69 +41,45 @@ def load(self, request):
...
+                except ObjectDoesNotExist:
+                    yield (row[1], row[2], row[0], rejected, False)
+
+    
+        failed_rows = Assignment.update_assignments(generate_assignments(reader))
+        if failed_rows:
             messages.error(request, 
-                'Assignments aborted.\nThese are the rows that failed with identification of the cells at fault:\n' + ', \n'.join(failed_rows))
-            return HttpResponseRedirect(reverse('admin:core_assignment_changelist'))
+                'Assignments aborted. These are the rows that failed with identification of the cells at fault: ' + ' | '.join(failed_rows))
```
tdowds: Keep it simple and just say "Assignment upload aborted. These assignments failed: "

huxley/core/models.py
```diff
@@ -273,28 +275,54 @@ def add(committee, country, school, rejected):
...
+                    # so that its delegates are deleted by cascade.
+                    remove(old_assignment)
+                    add(committee, country, school, rejected)
 
-        for old_assignment in assignment_dict.values():
-            remove(old_assignment)
+                del assignment_dict[key]
+            else:
+                if not Committee.objects.filter(name=committee).exists():
+                    committee = committee + ' - INVALID'
```
tdowds: Since we're checking for existence here, might be better to inform the user that that committee/country/school does not exist rather than that it's invalid

kmeht:
This is a great feature! It'll be really helpful to whomever uploads assignments.

I think the organization can be improved a little bit. It's currently difficult to understand. Some suggestions off the top of my head, in no particular order:

Before, we returned tuples of IDs from `generate_assignments`. But now, in the assignment code, we need the actual model objects to get their name. How about we actually return tuples of the models themselves from `generate_assignments`, instead of IDs?

It looks like we're catching `ObjectDoesNotExist` inside `generate_assignments`, which doesn't tell us which model specifically doesn't exist (or even if multiple don't exist!). Instead, how about we catch [each model's specific `DoesNotExist` exception](https://docs.djangoproject.com/en/1.9/ref/models/instances/#doesnotexist) inside `get_model` and then return some type of dummy (maybe just the name, instead of the object):
```python
def get_model(model, name, cache):
    if not name in cache:
        try:
            cache[name] = model.items.get(name=name)
        except model.DoesNotExist:
            cache[name] = name
    return cache[name]
```

With the adjustments below, the input tuples to `Assignment.update_assignments` will contain the model objects themselves (or just their names to indicate they don't exist), saving us the trouble of fetching them in that code.

---

Now, inside `Assignment.update_assignments`, one minor tip for code organization inside loops is to avoid `else` blocks. They're confusing (at least to me). Instead, make use of `if` and `continue`. Also, if you have nested `if` and `else`, consider negating the condition. For example:
```python
# BEFORE
for foo in whatever:
    if is_valid:
        something = foo.bar()
        if something:
            continue
        if something.else()
            ...
        do_this_regardless()
    else:
        invalid_stuff()

# AFTER
for foo in whatever:
    if not is_valid:
        invalid_stuff()
        continue
    
    something = foo.bar()
    if something:
        continue
    
    if something.else()
        ...
    do_this_regardless()
```

What do you think? Let's see how the code looks after these adjustments.

huxley/core/admin/assignment.py
```diff
@@ -4,9 +4,10 @@
 import csv
 
 from django.conf.urls import patterns, url
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.core.urlresolvers import reverse
 from django.http import HttpResponse, HttpResponseRedirect
+from django.core.exceptions import ObjectDoesNotExist
```
kmeht: nit: alphabetize

huxley/core/models.py
```diff
@@ -266,53 +266,56 @@ def update_assignments(cls, new_assignments):
...
-            is_valid = type(committee) is int
-
             key = (committee, country)
             if key in assigned:
                 # Make sure that the same committee/country pair is not being
                 # given to more than one school in the upload
-                committee = str(Committee.objects.get(pk=committee).name) + ' - DUPLICATE ASSIGNMENT IN CSV'
-                country = str(Country.objects.get(pk=country).name) + ' - DUPLICATE ASSIGNMENT IN CSV'
-                school = str(School.objects.get(pk=school).name)
+                committee = str(committee.name) + ' - DUPLICATE ASSIGNMENT IN CSV'
```
tdowds: So this output ends up looking something like this: 
```
('Trevor School', 'WHO - DUPLICATE ASSIGNMENT IN CSV', 'United States - DUPLICATE ASSIGNMENT IN CSV')
```
I think it would be more informative to just say: 
```
('Trevor School', 'WHO', 'United States') - DUPLICATE ASSIGNMENT IN CSV
```

huxley/core/models.py
```diff
@@ -266,53 +266,56 @@ def update_assignments(cls, new_assignments):
...
                 continue
 
-            if is_valid:
-                assigned.add(key)
-                old_assignment = assignment_dict.get(key)
+            # If the assignemnt contains no bad cells, then each value should
+            # have the type of its corresponding model.
+            is_valid = True
+            if type(committee) is not Committee:
+                committee = Committee(name=str(committee)+' - DOES NOT EXIST')
```
tdowds: I don't think we need to create a new Committee here. You can just do:
```
committee = str(committee) + ' - DOES NOT EXIST'
```

huxley/core/models.py
```diff
@@ -266,53 +266,56 @@ def update_assignments(cls, new_assignments):
...
+                is_valid = False
+            if type(country) is not Country:
+                country = Country(name=str(country)+' - DOES NOT EXIST')
```
tdowds: Same as above but this time with `Country`

huxley/core/models.py
```diff
@@ -266,53 +266,56 @@ def update_assignments(cls, new_assignments):
...
+                is_valid = False
+            if type(school) is not School:
+                school = School(name=str(school)+' - DOES NOT EXIST')
```
tdowds: Same as above but this time with `School`

huxley/core/models.py
```diff
@@ -266,53 +266,56 @@ def update_assignments(cls, new_assignments):
...
             key = (committee, country)
             if key in assigned:
                 # Make sure that the same committee/country pair is not being
                 # given to more than one school in the upload
-                committee = str(Committee.objects.get(pk=committee).name) + ' - DUPLICATE ASSIGNMENT IN CSV'
-                country = str(Country.objects.get(pk=country).name) + ' - DUPLICATE ASSIGNMENT IN CSV'
-                school = str(School.objects.get(pk=school).name)
+                committee = str(committee.name) + ' - DUPLICATE ASSIGNMENT IN CSV'
+                country = str(country.name) + ' - DUPLICATE ASSIGNMENT IN CSV'
+                school = str(school.name)
```
tdowds: This line will fail if you try assigning an assignment that has already been assigned to a school that doesn't exist.

huxley/core/models.py
m-j-mcdonald: I did this to improve style consistency in the output. If a Committee is created then I can use the name field without checking the type of the committee, i.e. otherwise to keep the output in a consistent format I would have to convert the committee to a string before saving it which would clutter the code a lot more.

huxley/core/admin/assignment.py
```diff
@@ -41,29 +42,44 @@ def load(self, request):
         reader = csv.reader(assignments['csv'])
 
         def get_model(model, name, cache):
-            if not name in cache:
-                cache[name] = model.objects.get(name=name)
+            try:
```
kmeht: We only want to try to fetch the model if it's not in the cache already.
```python
if not name in cache:
    try:
        ...
    except model.DoesNotExist:
        ...
return cache[name]
```

huxley/core/admin/assignment.py
```diff
@@ -41,29 +42,44 @@ def load(self, request):
...
+                if not name in cache:
+                    cache[name] = model.objects.get(name=name)
+            except ObjectDoesNotExist:
```
kmeht: This is very minor, since we're only executing one fetch in the `try` block, but each model has its own `DoesNotExist` exception type. I used it in the example above. I wouldn't comment ordinarily, but while you're making the above change, sneak this in there :)

huxley/core/admin/assignment.py
```diff
@@ -41,29 +42,44 @@ def load(self, request):
...
+                committee = get_model(Committee, row[1], committees)
+                country = get_model(Country, row[2], countries)
+                school = get_model(School, row[0], schools)
+                yield (committee, country, school, rejected)
+
+    
+        failed_rows = Assignment.update_assignments(generate_assignments(reader))
+        if failed_rows:
+            messages.error(request, 
+                'Assignment upload aborted. These assignments failed: ' + ' | '.join(failed_rows))
```
kmeht: Maybe we can join with a newline instead of a pipe? Based on the screenshot, it might be overwhelming if there are a lot of errors.

kmeht:
@m-j-mcdonald, is there a reason everything is wrapped in `str()`? Seems like we use that all over the place, even when fields are strings like model names. Is it because of unicode/byte string incompatibilities in Python?

huxley/core/models.py
```diff
@@ -283,7 +283,7 @@ def remove(assignment_data):
                 # given to more than one school in the upload
                 committee = str(committee.name)
                 country = str(country.name)
-                failed_assignments.append(str(committee,country) + ' - DUPLICATE ASSIGNMENT IN CSV')
+                failed_assignments.append(str((committee, country)) + ' - DUPLICATE ASSIGNMENT IN CSV')
```
tdowds: I still think this should be something like this:
```
('WHO', 'United States') - ASSIGNED TO MORE THAN ONE SCHOOL
```

m-j-mcdonald:
@kmeht The use of `str()` was to get a more readable format in the output. The name fields of the models, such as committee.name, are represented as u'NAME' otherwise (which I'm guessing has to do with them being retrieved from the database but I'm new to django). Putting things in str() made it easier to read for the end user.

m-j-mcdonald:
New screenshot:
screen shot 2016-06-15 at 1 02 57 pm

kmeht:
Ok, so it does have to do with unicode and byte strings. There's probably some nicer way to do it, but not worth blocking the PR over. @tdowds, if you have no further comments, merge it!

tdowds:
closed via 774a7540ec60fe690d5d2929eac23717176f94aa

PR #471, opened 2016-06-17
Close Registration From Admin Panel

nathanielparke:
Allows for external to close registration from the admin panel and transitions to the use of the conference model as SSOT for all conference registration state data.

### DIFF
--- a/huxley/api/views/user.py
+++ b/huxley/api/views/user.py
@@ -15,7 +15,7 @@ from huxley.accounts.models import User
 from huxley.accounts.exceptions import AuthenticationError, PasswordChangeFailed
 from huxley.api.permissions import IsPostOrSuperuserOnly, IsUserOrSuperuser
 from huxley.api.serializers import CreateUserSerializer, UserSerializer
-from huxley.core.models import School
+from huxley.core.models import School, Conference
 
 
 class UserList(generics.ListCreateAPIView):
@@ -24,7 +24,7 @@ class UserList(generics.ListCreateAPIView):
     permission_classes = (IsPostOrSuperuserOnly,)
 
     def create(self, request, *args, **kwargs):
-        if settings.REGISTRATION_OPEN:
+        if Conference.get_current().open_reg:
             return super(UserList, self).create(request, *args, **kwargs)
         raise PermissionDenied('Conference registration is closed.')
 
--- a/huxley/api/views/user.py
+++ b/huxley/api/views/user.py
@@ -15,7 +15,7 @@ from huxley.accounts.models import User
 from huxley.accounts.exceptions import AuthenticationError, PasswordChangeFailed
 from huxley.api.permissions import IsPostOrSuperuserOnly, IsUserOrSuperuser
 from huxley.api.serializers import CreateUserSerializer, UserSerializer
-from huxley.core.models import School, Conference
+from huxley.core.models import Conference, School
 
 
 class UserList(generics.ListCreateAPIView):
--- a/huxley/settings/registration.py
+++ b/huxley/settings/registration.py
@@ -1,4 +1,0 @@
-# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
-# Use of this source code is governed by a BSD License (see LICENSE).
-
-REGISTRATION_OPEN = True
--- a/huxley/www/static/js/huxley.browserify.js
+++ b/huxley/www/static/js/huxley.browserify.js
@@ -15,6 +15,7 @@ var Huxley = require('./huxley/components/Huxley');
 
 var AdvisorAssignmentsView = require('./huxley/components/AdvisorAssignmentsView');
 var AdvisorProfileView = require('./huxley/components/AdvisorProfileView');
+var ConferenceContext
 var ForgotPasswordView = require('./huxley/components/ForgotPasswordView');
 var LoginView = require('./huxley/components/LoginView');
 var NotFoundView = require('./huxley/components/NotFoundView');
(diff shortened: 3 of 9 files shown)


huxley/www/static/js/huxley.browserify.js
```diff
@@ -35,9 +35,10 @@ var routes = (
     <Route path="/login" handler={LoginView} />
     <Route path="/password" handler={ForgotPasswordView} />
     <Route path="/password/reset" handler={PasswordResetSuccessView} />
-    <Route path="/register" handler={RegistrationView} />
+    <Route path="/register/open" handler={RegistrationView} />
```
kmeht: How about using the conference context here to conditionally assign the handler, so there's actually no way to get to the registration view if it's closed.

huxley/www/static/js/huxley/components/ConferenceContext.js
```diff
@@ -21,7 +21,9 @@ var ConferenceContext = {
     }),
     external: React.PropTypes.string,
     registration_fee: React.PropTypes.number,
-    delegate_fee: React.PropTypes.number
+    delegate_fee: React.PropTypes.number,
+    registration_open: React.PropTypes.bool,
+    registration_waitlist:React.PropTypes.bool
```
kmeht: nit: space after colon, trailing comma,

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -89,6 +89,14 @@ var RegistrationView = React.createClass({
     };
   },
 
+  componentWillMount: function() {
```
kmeht: Once the registration handler is set to either open, closed, or waitlist, we won't need this function any more.

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -89,6 +89,14 @@ var RegistrationView = React.createClass({
...
+    if (!conference.registration_open) {
```
kmeht: Also, shouldn't this be `this.context.conference` anyway?

tdowds: Is there a reason why we're putting this logic here instead of on the Register button in the LoginView. We should condition on `conference.registration_open` such that if it's true the `href` on the Register button is `register/open` and if it's false the href is `register/closed`. We should do something similar for waitlisting schools by conditioning on `conference.registration_waitlist` such that if its true in the onSuccess function for the ajax call in RegistrationView it transitions to `register/waitlist` or `register/success` otherwise.

nathanielparke: @tdowds Yeah so if a user navigates to "/register" not using the UI then they will be able to see the registration page which is something we don't want. 

@kmeht Initially didn't implement it like that because of the file structure but I will take your suggestion as it is cleaner

tdowds:
Might also be worth updating the current settings folder to either not use REGISTRATION_OPEN or to set it equal to the current conference's `registration_open` variable

nathanielparke:
@tdowds Yeah I think removing that setting is the right way to go.

huxley/www/static/js/huxley.browserify.js
```diff
@@ -15,13 +15,14 @@ var Huxley = require('./huxley/components/Huxley');
 
 var AdvisorAssignmentsView = require('./huxley/components/AdvisorAssignmentsView');
 var AdvisorProfileView = require('./huxley/components/AdvisorProfileView');
+var ConferenceContext
```
kmeht: wat

huxley/www/static/js/huxley.browserify.js
```diff
@@ -35,7 +36,9 @@ var routes = (
     <Route path="/login" handler={LoginView} />
     <Route path="/password" handler={ForgotPasswordView} />
     <Route path="/password/reset" handler={PasswordResetSuccessView} />
-    <Route path="/register" handler={RegistrationView} />
+    <Route path="/register" handler={ global.conference.registration_open ?
```
kmeht: style nit:
```
<Route
  path="/register"
  handler={global.conference.registraton_open
    ? RegistrationView
    : RegistrationClosedView
  }
/>

**Comment by kmeht** [comment]
LGTM, ready to merge after couple of fixes

File: huxley/settings/registration.py [hunk 7]
```diff
@@ -1,4 +0,0 @@
-# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
-# Use of this source code is governed by a BSD License (see LICENSE).
-
-REGISTRATION_OPEN = True
```
kmeht: Search the codebase for `REGISTRATION_OPEN`, it's used in a few places.

nathanielparke: Interesting. Tests pass. This mean missing test coverage.

tdowds: Did you remove the references in `huxley/api/views/user.py` and in the tests in `huxley/api/tests/test_user.py` ?

nathanielparke: See latest

huxley/api/views/user.py
```diff
@@ -15,7 +15,7 @@
 from huxley.accounts.exceptions import AuthenticationError, PasswordChangeFailed
 from huxley.api.permissions import IsPostOrSuperuserOnly, IsUserOrSuperuser
 from huxley.api.serializers import CreateUserSerializer, UserSerializer
-from huxley.core.models import School
+from huxley.core.models import School, Conference
```
tdowds: nit: alphabetize

tdowds:
Assuming the app runs and this feature works when you test it manually, I think it's good to merge

PR #473, opened 2016-06-20
Logging updates

tdowds:
Conditions logging behavior based on whether were in prod or dev. If in dev logs are printed to the console and logged to logging. In prod, logs are stored in prod. 

Updated exception handler to log the stacktrace to logging and email the admin. Also prints to the console if in dev.


huxley/core/middlewares.py
```diff
@@ -1,24 +1,34 @@
 # Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
 # Use of this source code is governed by a BSD License (see LICENSE).
 
-import json, logging
+import json, logging, sys, traceback
+
+from huxley.settings.main import DEBUG
```
kmeht:
```python
from django.conf import settings
```
and then `settings.DEBUG` in code: https://docs.djangoproject.com/en/1.8/topics/settings/#using-settings-in-python-code

huxley/core/middlewares.py
```diff
@@ -1,24 +1,34 @@
...
-        logger = logging.getLogger('huxley')
-        logger.exception(exception)
 
-class ServerLoggingMiddleware(object):
     def process_exception(self, request, exception):
-        logger = logging.getLogger('huxley.server')
-        logger.exception(exception)
+        logger = logging.getLogger('huxley.exception')
+        exc_type, exc_value, exc_traceback = sys.exc_info()
+        exc_traceback = ''.join(traceback.format_exception(exc_type, exc_value, exc_traceback))
```
kmeht: Should we join on newline? Otherwise the stack trace would be hard to read as a big blob.

tdowds: It looks good in the admin panel, but in the console it's just kind of a hard to read blob. That's why I put the call to `logger.exception(exception)` at the bottom

huxley/settings/logging.py
```diff
@@ -1,12 +1,14 @@
 # Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
 # Use of this source code is governed by a BSD License (see LICENSE).
 
+from .main import DEBUG
```
kmeht: This may not work if the local settings overrides the value of `DEBUG`. That is to say, I believe this value is always going to be `True`. Not sure how to get around that, I'll think about it.

huxley/core/middlewares.py
```diff
@@ -1,24 +1,34 @@
...
+        log = json.dumps({
+              'message': exc_traceback,
+              'uri': request.path,
+              'status_code': 500,
+              'username': request.user.username})
+        logger.error(log)
```
kmeht: I don't think I understand the logger fully. Why do we pass in a JSON string into `logger.error`, but below, we pass an Exception instance to `logger.exception`? Are their implementations different?

tdowds: I believe so. `logger.error` eventually calls its `emit` method which we have overwritten in the DatabaseHandler to expect a json string that it parses in order to create a LogEntry object. I don't think `logger.exception` calls `emit` but instead gets the exceptions traceback and prints it to the console.

huxley/logging/LoggingEmailBackend.py
```diff
@@ -0,0 +1,42 @@
...
+            for email in email_messages:
+                recipients = email.to[0] if len(email.to) == 1 else ', '.join(email.to)
+                log = json.dumps({
+                  'message': "Sending email",
+                  'uri': str(recipients),
+                  'status_code': 0,
+                  'username': ''})
+                logger.info(log)
+
+                return super(LoggingEmailBackend, self)._send(email)
```
tdowds: I don't think the current exception logger will log an exception raised from trying to send an email. So in order for emails to not fail silently we can put a try catch around this and catch an `SMTPException` and log it and send it the admin.

tdowds:
It seems pretty standard to use a different email backend for development and another one for prod which is why I created two different classes.

https://docs.djangoproject.com/en/1.8/topics/email/#dummy-backend

Also for reference
http://stackoverflow.com/questions/7552283/how-can-i-log-all-outgoing-email-in-django

huxley/settings/main.py
```diff
@@ -30,7 +30,7 @@
 TIME_ZONE = 'America/Los_Angeles'
 USE_I18N = False
 
-EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+EMAIL_BACKEND = 'huxley.logging.LoggingEmailBackend.DevLoggingEmailBackend'
```
kmeht: We'll want to update the value in the local.py.default file (and we need to update it in our deployments also).

huxley/logging/LoggingEmailBackend.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
```
kmeht: Let's name this file `mail` to be consistent with Django's naming structure.

huxley/logging/LoggingEmailBackend.py
```diff
@@ -0,0 +1,42 @@
...
+
+from smtplib import SMTPException
+
+
+class DevLoggingEmailBackend(dummy.EmailBackend):
+
+    def send_messages(self, email_messages):
+        logger = logging.getLogger('huxley.api')
+        for email in email_messages:
+            recipients = email.to[0] if len(email.to) == 1 else ', '.join(email.to)
```
kmeht: Just the join will work, it won't add the delimiter if the length of the list is just one:
```python
recipients = ', '.join(email.to)
```

huxley/logging/LoggingEmailBackend.py
```diff
@@ -0,0 +1,42 @@
...
+            log = json.dumps({
+              'message': "Sending email",
+              'uri': str(recipients),
```
kmeht: Do we need the `str()`? Doesn't `join` already return a string?

huxley/logging/LoggingEmailBackend.py
```diff
@@ -0,0 +1,42 @@
...
+
+class LoggingEmailBackend(smtp.EmailBackend):
+
+    def send_messages(self, email_messages):
+        logger = logging.getLogger('huxley.api')
+        exc_logger = logging.getLogger('huxley.exception')
+
+        with transaction.atomic():
+            for email in email_messages:
+                recipients = email.to[0] if len(email.to) == 1 else ', '.join(email.to)
```
kmeht: Let's extract this logging into a module-level function and use it in both backends:
```python
def log_email(email):
    recipients = ...
    logger.info(json.dumps({
        ...,
        ...,
    }))
```

kmeht:
LGTM, a couple of comments and it's ready to merge.

tdowds:
Updated

tdowds:
closed via 3d94ccbb26c72c4a77db14505c713ab2bf329504

PR #476, opened 2016-06-22
Update OSX.md

tdowds:


nathanielparke:
LGTM

docs/setup/OSX.md
```diff
@@ -47,21 +47,35 @@ Next, configure your author information. Use your name and email as they appear
...
     $ python manage.py migrate
-    $ python manage.py loaddata countries committees advisor chair
+    $ python manage.py loaddata countries committees advisor chair conference
 
-We use Pipeline to collect, compile, and compress our static assets. It simply hooks into Django's `collectstatic` management command:
+Finally, we use Pipeline to collect, compile, and compress our static assets. It simply hooks into Django's `collectstatic` management command:
 
     $ python manage.py collectstatic --noinput
+    
+If you run into an error here run `ulimit -n 512`.
```
nathanielparke: Explanation on why we use this might be good for new devs. Also Maybe suggest adding it to their bash profile.

docs/setup/OSX.md
```diff
@@ -47,21 +47,35 @@ Next, configure your author information. Use your name and email as they appear
...
 
     $ python manage.py runserver
+    
+If you haven't created a superuser yet, you should do that now:
+
+    $ python manage.py createsuperuser
+    
+Your superuser credentials is what will allow you to login to the admin panel at `localhost:8000/admin/`
 
 With that, you're ready to go; start hacking!
```
nathanielparke: You can also suggest they check out all the django management commands by using

`python manage.py`

alone. It lists out the all the possible commands for them to run. Knowing those is really helpful to keep your own development environment in check.

Issue #353, opened 2014-10-29
Don't fail emails silently


kmeht:
Currently, to avoid borking the request, we have `fail_silently=True` in `send_mail` calls. This is bad because we don't know if the emails are working or not. Instead, we should catch the exception and log or do something so that the tech officer or external officer know what's going on.

More context: https://docs.djangoproject.com/en/dev/topics/email/#send-mail

kmeht:
@tdowds, assigning to you since we have some rudimentary logging set up now :)

tdowds:
Thanks man

kmeht:
Making this high-pri; we should really log this.

tdowds:
Closed via 3b8cab0209f4f0aa8f6ff8ac85aee52b5b3743c5

PR #477, opened 2016-06-27
#353 does not fail emails silently.

tdowds:
Should we only catch SMTPExceptions or should we catch exceptions more generically in case emails fail for another reason (like a bug in our own code or something like that)? Also not having the failure be emailed to tech in case the exception has to do with something on our side not allowing us to send emails. In that case we would get another exception when we failed to send the email to tech. However that email would fail for the same reason causing yet another exception to be generated, logged and attempted to be emailed to tech, but it would fail as well. This behavior would continue until we hit the bottom of the stack. To avoid this behavior I'm just having it log and not email.

### DIFF
--- a/huxley/logging/mail.py
+++ b/huxley/logging/mail.py
@@ -6,6 +6,8 @@ import json, logging
 from django.core.mail.backends import dummy, smtp
 from django.db import transaction
 
+from smtplib import SMTPException
+
 
 class DevLoggingEmailBackend(dummy.EmailBackend):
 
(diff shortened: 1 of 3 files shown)


huxley/logging/mail.py
```diff
@@ -21,7 +23,18 @@ def send_messages(self, email_messages):
...
+                    return super(LoggingEmailBackend, self)._send(email)
+                except SMTPException:
+                    logger = logging.getLogger('huxley.api')
+                    exc_type, exc_value, exc_traceback = sys.exc_info()
+                    exc_traceback = ''.join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+                    log = json.dumps({
+                          'message': exc_traceback,
+                          'uri': ', '.join(email.to),
+                          'status_code': 500,
+                          'username': ''})
```
kmeht: Is there any way we'd have access to the current user in this method? Would be useful information to have.

huxley/logging/mail.py
```diff
@@ -21,7 +23,18 @@ def send_messages(self, email_messages):
...
+                    logger.error(log)
```
kmeht: Do we also want to do a `logger.exception` call, like in the exception logger?

kmeht:
LGTM, two comments before merging.

tdowds:
Closed via 3b8cab0209f4f0aa8f6ff8ac85aee52b5b3743c5

Issue #475, opened 2016-06-20
Make writing tests suck less: AutoTestCase


kmeht:
Currently, writing tests for our API views is a big pain in the ass. They're long, and mostly boilerplate.

Could we take advantage of Python's dynamism and write an AutoTestCase? Something that, given some classes (API URL, Serializer, Model, whatever), could iterate over all the fields to figure out what they could be, and then send various requests and verify the responses?

This would be a MASSIVE win for developer efficiency and code quality: writing tests would be so much easier, so we'd write more tests and get better coverage.

We'd also have to write an AutoTestCaseTestCase to make sure that AutoTestCase is working properly (who watches the watchmen, etc.) 😃 

@nathanielparke, @tdowds: sounds like a task for a relatively experienced engineer. Any interest?

kmeht:
I'll take a stab at this. No guarantees on completion date though.

PR #482, opened 2016-07-02
Introduce RetrieveAPIAutoTestCase

kmeht:
This introduces the first auto test case for the retrieve API. The implementor provides the actual object under test, and a tuple of (username, password, expected_error). The expected error is what that user should get from the API. If it's `None`, that means the request should be accepted.

If the request is accepted, we then use the serializer to figure out what fields the object is going to have. Most of the time, it's pretty straightforward, but sometimes (like for decimal fields, or relational fields), we have to do some transformations.

This will (hopefully) make it much easier to add test cases to the API. Or, at the very least, it'll allow us to clean up our fugly existing test cases.

Task: #475


huxley/api/tests/auto.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2011-2016 Berkeley Model United Nations. All rights reserved.
+# Use of this source code is governed by a BSD License (see LICENSE).
+
+from django.db import models
+from rest_framework import serializers
+
+from huxley.api.tests import RetrieveAPITestCase
+
+
+class RetrieveAPIAutoTestCase(RetrieveAPITestCase):
```
tdowds: As we add more AutoTestCases, do you think we will be able to reuse some of the functionality by making an AbstractAutoTest case or something along those lines?

tdowds:
This is awesome!

huxley/api/tests/auto.py
kmeht: I hope so! I figure we'll take it one step at a time, and the requirements of the other AutoTestCases will shape what the abstract one will look like.

huxley/api/tests/school/test_school.py
```diff
@@ -1,120 +1,33 @@
...
-        '''it should allow a get request from a superuser.'''
-        school = TestSchools.new_school()
+    @classmethod
+    def get_users(cls, test_object):
         TestUsers.new_superuser(username='user1', password='user1')
+        return (
+            (None, None, cls.NOT_AUTHENTICATED),
+            (test_object.advisor.username, 'test', None),
+            ('user1', 'user1', None),
+        )
```
nathanielparke: I don't have a better way to write this now, but this return statement is not very clear. Maybe switching over to a python dictionary would allow more clarity and readability. For right now I think a class comment would suffice so that devs trying to replicate can easily do so.

nathanielparke:
This is a great start. As mentioned above I think readability will be important as we introduce more automation. I think there might need to be a bit more abstraction added between the implemented test case and the factory. That decoupling can come after we get rid of the old test though. LGTM

huxley/api/tests/school/test_school.py
kmeht: Yeah, agreed that this is pretty terrible. Let's keep thinking on it and make improvements later.

PR #483, opened 2016-07-02
Fix Spanish speaking & Chinese speaking delegate issue

m-j-mcdonald:
This modifies the SchoolSerializer's validate method to collect all validation issues into a single ValidationError with a dict of every field checked by the method with an issue. It introduces a check on whether the # of Spanish or Chinese speaking delegates exceeds the total # of delegates, as well as fix a bug where an invalid phone number, e.g. (123) 456 78, would raise a ValidationError but the error would not appear in registration view.

### DIFF
--- a/huxley/api/serializers/school.py
+++ b/huxley/api/serializers/school.py
@@ -81,7 +81,6 @@ class SchoolSerializer(serializers.ModelSerializer):
         advanced_delegates = data.get('advanced_delegates')
         spanish_speaking_delegates = data.get('spanish_speaking_delegates')
         chinese_speaking_delegates = data.get('chinese_speaking_delegates')
-        total_delegates = beginner_delegates + intermediate_delegates + advanced_delegates
 
         if primary_phone:
             try:
@@ -99,6 +98,14 @@ class SchoolSerializer(serializers.ModelSerializer):
                     validators.phone_domestic(secondary_phone)
             except serializers.ValidationError:
                 invalid_fields['secondary_phone'] = 'This is an invalid phone number.'
+
+        total_delegates = 0
+        if beginner_delegates:
+            total_delegates += beginner_delegates
+        if intermediate_delegates:
+            total_delegates += intermediate_delegates
+        if advanced_delegates:
+            total_delegates += advanced_delegates
 
         if spanish_speaking_delegates > total_delegates:
             invalid_fields['spanish_speaking_delegates'] = 'Cannot exceed total delegates'
--- a/huxley/api/serializers/school.py
+++ b/huxley/api/serializers/school.py
@@ -72,20 +72,42 @@ class SchoolSerializer(serializers.ModelSerializer):
 
 
     def validate(self, data):
+        invalid_fields = {}
         international = data.get('international')
         primary_phone = data.get('primary_phone')
         secondary_phone = data.get('secondary_phone')
+        beginner_delegates = data.get('beginner_delegates')
+        intermediate_delegates = data.get('intermediate_delegates')
+        advanced_delegates = data.get('advanced_delegates')
+        spanish_speaking_delegates = data.get('spanish_speaking_delegates')
+        chinese_speaking_delegates = data.get('chinese_speaking_delegates')
+        total_delegates = beginner_delegates + intermediate_delegates + advanced_delegates
 
         if primary_phone:
-            if international:
-                validators.phone_international(primary_phone)
-            else:
-                validators.phone_domestic(primary_phone)
+            try:
+                if international:
+                    validators.phone_international(primary_phone)
+                else:
+                    validators.phone_domestic(primary_phone)
+            except ValidationError:
+                invalid_fields['Primary Phone'] = 'Invalid phone number'
         if secondary_phone:
-            if international:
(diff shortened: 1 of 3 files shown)


huxley/api/tests/__init__.py
```diff
@@ -97,7 +97,7 @@ def assertInvalidEmailFormat(self, response, field):
 
     def assertInvalidPhone(self, response, field):
         self.assertEqual(response.data, {
-            u'non_field_errors': [u'This is an invalid phone number.']})
+            '%s' % field: [u'This is an invalid phone number.']})
```
kmeht: The key can just be `field` directly, e.g.
```python
{field: [u'This...']}
```

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -109,7 +109,7 @@ var RegistrationView = React.createClass({
           onSubmit={this._handleSubmit}>
           <div>
             <h1>Register for Berkeley Model United Nations</h1>
-            <p>Please fill out the following information to register your school
+            <p>Pleasee fill out the following information to register your school
```
kmeht: Revert

huxley/api/serializers/school.py
```diff
@@ -74,20 +74,48 @@ class Meta:
...
         international = data.get('international')
         primary_phone = data.get('primary_phone')
         secondary_phone = data.get('secondary_phone')
+        beginner_delegates = data.get('beginner_delegates')
+        intermediate_delegates = data.get('intermediate_delegates')
+        advanced_delegates = data.get('advanced_delegates')
+        spanish_speaking_delegates = data.get('spanish_speaking_delegates')
+        chinese_speaking_delegates = data.get('chinese_speaking_delegates')
 
         if primary_phone:
```
kmeht: Shouldn't primary phone always be required?

huxley/api/serializers/school.py
```diff
@@ -74,20 +74,48 @@ class Meta:
...
-            if international:
-                validators.phone_international(primary_phone)
-            else:
-                validators.phone_domestic(primary_phone)
+            try:
```
kmeht: Seems like this can be extracted into a function, since we do the same thing twice.

huxley/api/serializers/school.py
```diff
@@ -74,20 +74,48 @@ class Meta:
...
+        total_delegates = 0
+        if beginner_delegates:
+            total_delegates += beginner_delegates
+        if intermediate_delegates:
+            total_delegates += intermediate_delegates
+        if advanced_delegates:
+            total_delegates += advanced_delegates
+
+        if spanish_speaking_delegates > total_delegates:
+            invalid_fields['spanish_speaking_delegates'] = 'Cannot exceed total delegates'
```
kmeht: Period at the end. Also, "total number of delegates."

huxley/api/serializers/school.py
```diff
@@ -74,20 +74,48 @@ class Meta:
...
-                validators.phone_domestic(secondary_phone)
+            try:
+                if international:
+                    validators.phone_international(secondary_phone)
+                else:
+                    validators.phone_domestic(secondary_phone)
+            except serializers.ValidationError:
+                invalid_fields['secondary_phone'] = 'This is an invalid phone number.'
+
+        total_delegates = 0
```
kmeht: Here's one way we can trim this down:
```python
total_delegates = sum((
    beginner_delegates or 0,
    intermediate_delegates or 0,
    advanced_delegates or 0,
))
```

huxley/www/static/js/huxley/components/RegistrationView.js
```diff
@@ -367,6 +367,8 @@ var RegistrationView = React.createClass({
               <li>
                 <input
                   type="text"
+                  id="input_delegates"
```
kmeht: What do this `id` and `name` attributes do? They don't seem to be referenced anywhere else in this diff.

huxley/api/serializers/school.py
```diff
@@ -72,20 +72,45 @@ class Meta:
...
+
+        def validate_phone(phone, international):
             if international:
-                validators.phone_international(primary_phone)
+                validators.phone_international(phone)
             else:
-                validators.phone_domestic(primary_phone)
+                validators.phone_domestic(phone)
+
+        if primary_phone:
```
m-j-mcdonald: The removal of this line runs afoul of several unit tests, and it preexisted before this PR (removed line 79), so I left it in in case it serves a purpose I am unaware of. However, it does seem unnecessary since primary_phone is a required field.

huxley/api/tests/__init__.py
m-j-mcdonald: This is the formatting used elsewhere in the file so it was used for consistency.

huxley/api/serializers/school.py
kmeht: Ok, no problem. Let's investigate the cause of that separately.

huxley/api/tests/__init__.py
kmeht: You're right, consistency is important. I'll change these all in a later commit.

PR #484, opened 2016-07-03
AutoTest: Introduce User named tuple.

kmeht:
(#475) The tuples used to specify users and expected errors were really confusing. We can use named tuples to make this more clear. We can also add some hackery so that the fields are optional (and set to None by default).

### DIFF
--- a/huxley/api/tests/auto.py
+++ b/huxley/api/tests/auto.py
@@ -1,14 +1,22 @@
 # Copyright (c) 2011-2016 Berkeley Model United Nations. All rights reserved.
 # Use of this source code is governed by a BSD License (see LICENSE).
 
+from collections import namedtuple
+
 from django.db import models
 from rest_framework import serializers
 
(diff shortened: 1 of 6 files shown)


huxley/api/tests/auto.py
```diff
@@ -1,14 +1,22 @@
...
 from rest_framework import serializers
 
 from huxley.api.tests import RetrieveAPITestCase
 
 
+User = namedtuple('User', ['username', 'password', 'expected_error'])
+User.__new__.__defaults__ = (None, None, None)
+
+
+EXP_NOT_AUTHENTICATED = 'exp_not_authenticated'
```
tdowds: As we add more errors, it might be a good idea to wrap this in a dictionary and call it `expected_errors` just so it's clearer we're referencing errors.

kmeht: Noted! 👍

Issue #480, opened 2016-07-01
AdvisorAssignmentView updates for delegate names


tdowds:
After advisors finalize their assignments, a new column should appear on the AdvisorAssignmentView where advisors can select delegates they want to assign to that assignment via a dropdown

When selecting delegates Advisors should not be allowed to assign one delegate to a double delegate assignment. Additionally, advisors should not be able to assign a delegate to more than one assignment.

tdowds:
Closed via 4ad41c27f04750b4603f24f01440b760d29f8bfa

Issue #463, opened 2016-05-23
Make School Registration Information Mutable


nathanielparke:
Currently the AdvisorProfileView does not allow for advisors to edit the data that they registered with. This should be changed. We should create a button at the top of the data view that allows for an advisor to edit the data then save or discard the edit. In the backend we should also add a tag that signifies when the data was last updated. This is so that External will know if their information for a given school is out of date.

tdowds:
@michaeleliot if memory serves me right, the way we use to do this was whenever the cursor hovered over one of the fields it would highlight. Then you could click on the field to update the text. Then you would hit a save button to submit those changes. So one possible approach would be to take it in these chunks: 

- [ ] Highlight table cell on hover
- [ ] Clicking a cell creates a textbox in which to edit the information
- [ ] Changing focus away from the cell restores it to its normal state
- [ ] Add save button
- [ ] Make a request to our API to save those changes

nathanielparke:
@michaeleliot Per discussion with @tdowds we realized that editing data of a school can cause some logistical problems for external. The issue is that when an advisor changes their info if we silently allow them to change the data and don't notify external then there is a data inconsistency between what data external is using and the data that is in our DB. 

Potential fixes to this:

1. Change log for external to review when accessing the admin panel
2. Send a [Insert time length here] digest email to external of all the data that was changed
3. Send an email to external every time data is changed

Feel free to PM myself or Trevor with questions.

kmeht:
@nathanielparke, why does external use a different version of school data than our DB? Isn't that the whole problem Huxley was supposed to solve? Is there a way we can solve this deeper problem?

tdowds:
Discussed this with external here are the specifications for this task:
Fields enabled to be changed:
* Advisor Information
 * First Name
 * Last Name
* School Information
 * Address
 * City
 * Zip
* Primary Contact Information & Secondary Contact Information
 * Name
 * Email
 * Phone

Like @nathanielparke said, external needs to be notified if there's a change, therefore on a successful change we're going to send external an email.

@michaeleliot this task is good to go if you want to start working on it (I had told him to hold off cause external was hesitant about doing this).

@kmeht I believe external downloads data from Huxley but if a change is made without external knowing, she could be using inconsistent information (like an incorrect email).

nathanielparke:
fixed in 1bcb28149f503c7f8fc212ba120ace94b96823b9

PR #485, opened 2016-07-05
Delegate name finalization functionality (#480)

tdowds:
See #480 for context

For what's still left to do we need to:

1. Load the delegates for that particular school from the DelegateStore in #468 
2. Render those delegates in drop downs on the UI
 * To do this it might be worth abstracting the current CountrySelect component into a general Select component because we'll want to reuse the functionality of not selecting an option more than once.
 * Assuming this first diff is good, that will be my next focus
3. Make a request to the Delegate API to assign assignments to each delegate
 * Might be good to do some validation here as well to ensure no delegate is assigned to more than one assignment and that a double delegate assignment has exactly two delegates

### DIFF
--- a/huxley/api/serializers/school.py
+++ b/huxley/api/serializers/school.py
@@ -10,6 +10,7 @@ class SchoolSerializer(serializers.ModelSerializer):
     fees_owed = serializers.FloatField(read_only=True)
     fees_paid = serializers.FloatField(read_only=True)
     assignments_finalized = serializers.BooleanField(required=False)
+    delegate_names_finalized = serializers.BooleanField(required=False)
     countrypreferences = serializers.ListField(child=serializers.IntegerField(required=False), source='country_preference_ids')
     committeepreferences = serializers.PrimaryKeyRelatedField(allow_empty=True, many=True, queryset=Committee.objects.all(), required=False)
 
@@ -49,6 +50,7 @@ class SchoolSerializer(serializers.ModelSerializer):
             'fees_owed',
             'fees_paid',
             'assignments_finalized',
+            'delegate_names_finalized',
         )
         extra_kwargs = {
         'committeepreferences': {'required': False},
--- a/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
+++ b/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
@@ -64,7 +64,8 @@ var AdvisorAssignmentsView = React.createClass({
   },
 
   render: function() {
-    var finalized = CurrentUserStore.getFinalized();
+    var assignmentsFinalized = CurrentUserStore.getAssignmentFinalized();
+    var delegateNamesFinalized = CurrentUserStore.getDelegateNameFinalized();
     var conference = this.context.conference;
     return (
       <InnerView>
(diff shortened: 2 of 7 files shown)


huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -163,6 +187,35 @@ var AdvisorAssignmentsView = React.createClass({
...
+
+    On a succesful response back, we then make another call to the school api to
+    update the school to know its names have been finalized.
+    */
+    var confirm = window.confirm("Are you sure you want to finalize your delegate names?");
+    var school = CurrentUserStore.getCurrentUser().school;
+    if (confirm) {
+      this.setState({loading: true});
+      $.ajax ({
+        type: 'PUT',
```
kmeht: This should be `PATCH`, since it's a partial update.

huxley/api/serializers/school.py
```diff
@@ -49,6 +50,7 @@ class Meta:
             'fees_owed',
             'fees_paid',
             'assignments_finalized',
+            'delegate_names_finalized',
```
kmeht: We should probably add some validation to the delegate update view to prevent it from being edited if the school's delegates are finalized. Currently, we have this boolean field but it doesn't do anything.

kmeht:
What's this UI supposed to look like? Will there be a separate page for advisors to add delegates, and then they will assign the delegates to assignments via a dropdown?

The pros to this are that they can have delegates that aren't assigned at all, kind of like "free agents." The cons are that there are multiple phases advisors have to go through, e.g. adding all the delegates and then hooking them up to assignments.

Do you think there's another way of building this UI that's better than the two-phase approach?

huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -163,6 +187,35 @@ var AdvisorAssignmentsView = React.createClass({
     }
   },
 
+  _handleFinalizeNames: function(event) {
+    /* Would appreciate some suggestions here. There are two things we need to do:
+      1. Update each delegate to have its assignment field set here.
```
kmeht: I think django rest framework has a concept of bulk editing. I imagine we'd send a bulk PATCH request to the delegate API and set all of their assignment_id fields to the relevant assignment.

tdowds:
So once the advisor hits Finalize Assignments the UI changes to look like this:
screen shot 2016-07-05 at 8 28 55 pm

Then after the advisor hits Finalize Delegate Names, the UI changes to look like this:
screen shot 2016-07-05 at 8 29 09 pm

I'm not really sure what other ways we can try to build the UI. The approach right now is that advisors will be able to add or remove delegates on the AdvisorRosterView until delegate names are finalized. However they can't assign a name to a particular assignment until they've finalized their assignments.

To me I think this is the most straightforward approach, but that doesn't necessarily mean it's the approach that's best suited for the advisors' (or external's) needs. When I spoke to external about this, they seemed to be on board with it. But I'm open to other suggestions.

nathanielparke:
I like this but again I am going to advocate for mutability of data. Similar to #463 I am of the mind that advisors know best what data they want in the system and it should be on us to find a solution to external not having the most recent data. I think what this means is that we need an effective notification system for external that notifies them via an email that data has changed. There needs to be some thought on how we want to do this though.

tdowds:
We're not going to be implementing delegate name finalization, for the good reason that advisors will probably want to make last minute changes. We'll need to build out away to alert external of those changes. #463 will be a good starting point for that. For now, going to close this PR.

PR #486, opened 2016-07-08
Updates React to 0.14.0

nathanielparke:
This PR updates our react dependency to 0.14.0. It adds a few react dependencies that are necessary due to the composition of the react project. See https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html

### DIFF
--- a/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
+++ b/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
@@ -80,7 +80,7 @@ var AdvisorAssignmentsView = React.createClass({
           <div className="tablemenu header" />
           <div className="table-container">
             <table className="table highlight-cells">
-              <tbody>
+              <thead>
                 <tr>
                   <th>Committee</th>
                   <th>Country</th>
@@ -90,6 +90,8 @@ var AdvisorAssignmentsView = React.createClass({
                     "Delete Assignments"}
                   </th>
                 </tr>
+              </thead>
+              <tbody>
                 {this.renderAssignmentRows()}
               </tbody>
             </table>
--- a/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
+++ b/huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
@@ -80,16 +80,18 @@ var AdvisorAssignmentsView = React.createClass({
           <div className="tablemenu header" />
           <div className="table-container">
             <table className="table highlight-cells">
-              <tr>
-                <th>Committee</th>
-                <th>Country</th>
-                <th>Delegation Size</th>
-                <th>{finalized ?
-                  "" :
-                  "Delete Assignments"}
-                </th>
-              </tr>
-              {this.renderAssignmentRows()}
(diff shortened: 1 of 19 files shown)


huxley/www/static/js/huxley/components/AdvisorAssignmentsView.js
```diff
@@ -80,16 +80,18 @@ var AdvisorAssignmentsView = React.createClass({
...
-                <th>Committee</th>
-                <th>Country</th>
-                <th>Delegation Size</th>
-                <th>{finalized ?
-                  "" :
-                  "Delete Assignments"}
-                </th>
-              </tr>
-              {this.renderAssignmentRows()}
+              <tbody>
```
tdowds: If this is wrapping around headers shouldn't we be using `<thead>`?

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -61,169 +62,171 @@ var AdvisorProfileView = React.createClass({
...
-                  {'$' + school.fees_paid.toFixed(2)}
-                </td>
-              </tr>
-              <tr>
-                <td className="fieldLabel">Balance</td>
-                <td className="field">
-                  {'$' + (school.fees_owed - school.fees_paid).toFixed(2)}
-                </td>
-              </tr>
+              <tbody>
```
tdowds: Same as thing, should we be using `<thead>` here and then `<tbody>` for everything else? Or does that affect the appearance too adversely?

huxley/www/static/js/huxley/components/LoginView.js
```diff
@@ -8,9 +8,12 @@
 var console = require('console');
 
 var $ = require('jquery');
-var React = require('react/addons');
+var classNames = require('classnames');
+var LinkedStateMixin = require('react-addons-linked-state-mixin');
+var React = require('react');
 var Router = require('react-router');
 
+
```
tdowds: I don't think there should be an extra space here, should there?

package.json
```diff
@@ -6,13 +6,17 @@
...
     "es6-promise": "2.0.0",
     "flux": "^2.1.1",
+    "fbjs": "0.8.3",
     "jquery": "2.1.1",
     "jquery-ui": "1.10.5",
     "js-cookie": "~2.0.3",
-    "react": "~0.13.3",
-    "react-router": "~0.13.0"
+    "react": "0.14.0",
+    "react-dom": "0.14.0",
```
tdowds: All of these dependencies are no longer part of the react package, you have to install them separately?

tdowds:
Thanks for doing this Nate, I could see this being really annoying to do

package.json
nathanielparke: yeah, React had a major restructuring. Read about it here: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html

kmeht:
LGTM. If I had one nit, (and this is a weird one), it would be to keep the `classnames` require as cx(). It's a way shorter name for a commonly used function. e.g.
```js
var cx = require('classnames');
```

PR #490, opened 2016-07-16
Upgrade to React v15.0.0

nathanielparke:
Upgrades react by removing the LinkedStateMixin

### DIFF
--- a/huxley/www/static/js/huxley/components/TextInput.js
+++ b/huxley/www/static/js/huxley/components/TextInput.js
@@ -27,7 +27,7 @@ var TextInput = React.createClass({
   },
 
   _handleChange: function(event) {
-    var value = event.target.value.replace(/[^\D]/, '');
+    var value = event.target.value.trim();
     this.props.onChange && this.props.onChange(value);
   },
 });
--- a/huxley/www/static/js/huxley/components/TextInput.js
+++ b/huxley/www/static/js/huxley/components/TextInput.js
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+ * Use of this source code is governed by a BSD License (see LICENSE).
+ */
+
+'use strict';
+
+var React = require('react')
+
+var TextInput = React.createClass({
+
+  propTypes: {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -13,10 +13,10 @@
     "jquery": "2.1.1",
     "jquery-ui": "1.10.5",
     "js-cookie": "~2.0.3",
-    "react": "0.14.8",
-    "react-dom": "0.14.8",
+    "react": "15.0.0",
+    "react-dom": "15.0.0",
     "react-router": "1.0.0",
-    "react-addons-linked-state-mixin": "0.14.8"
```
kmeht: We seemed to remove this everywhere, do we still need the dependency?

huxley/www/static/js/huxley/components/TextInput.js
```diff
@@ -0,0 +1,35 @@
...
+        type="text"
+        placeholder={this.props.placeholder}
+        onChange={this._handleChange}
+        value={this.props.value}
+      />
+    );
+  },
+
+  _handleChange: function(event) {
+    var value = event.target.value.replace(/[^\D]/, '');
```
kmeht: I don't think we should strip numbers. Text can certainly contain numbers, like addresses.

kmeht:
What's the criteria for deciding to use a custom change handler vs `_handleRawInputChange`?

tdowds:
I might have missed this somewhere, but is there a reason why we're removing LinkedStateMixin?

package.json
nathanielparke: Nope we can remove it. Will do that!

nathanielparke:
@kmeht Criteria was that _handleRawInputChange was the handler that is passed into our custom components such as NumberInput and TextInput. TextInput couldn't be used with things like addresses, passwords and usernames so the addition of those extra custom handlers was necessary. I think using _handleRawInputChange for all of them is a better idea though. Cleaner and matches linkedStateMixin functionality more closely.

kmeht:
@nathanielparke, you'll probably want to rebase on top of @m-j-mcdonald's latest diff, as there were some overlapping features.

PR #492, opened 2016-07-18
Make school registration information mutable

m-j-mcdonald:
This enables advisors to make changes to their registration information, as discussed in issue #463 . It adds a modified field, holding the date & time of the most recent modification, to the school model so external can check if they are using the most up-to-date information. The white boxes indicate fields that can be modified.
screen shot 2016-07-18 at 4 51 20 pm
screen shot 2016-07-18 at 4 44 14 pm


huxley/api/serializers/school.py
```diff
@@ -155,7 +156,7 @@ def validate_city(self, value):
 
         return value
 
-    def validate_zip(self, value):
+    def validate_zip_code(self, value):
```
m-j-mcdonald: This is unrelated to my issue, it was just a bug I noticed in the validator

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +325,105 @@ var AdvisorProfileView = React.createClass({
...
+        <label className="hint error">
+          {this.state.errors.school[field]}
+        </label>
+      );
+    }
+
+    return null;
+  },
+
+  _handleChange: function(fieldName, event) {
```
m-j-mcdonald: This is a general purpose handle function that could be used to clean up handlers elsewhere as well. The alternative to the conditional here would be to modify custom components (like PhoneInput) to pass the entire event object to the onChange prop instead of just a value.

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +325,105 @@ var AdvisorProfileView = React.createClass({
...
+    var change = {};
```
m-j-mcdonald: This is a general purpose handle function that could be used to clean up handlers elsewhere as well. The alternative to the conditional in the function body would be to modify custom components (like PhoneInput) to pass the entire event object to the onChange prop instead of just a value.

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +325,105 @@ var AdvisorProfileView = React.createClass({
...
+    // Handles both changes from HTMl and custom components
+    if (event.target) {
+      change[fieldName] = event.target.value;
+    } else {
+      change[fieldName] = event
+    }
+
+    this.setState(change);
+  },
+
```
m-j-mcdonald: This is a general purpose handle function that could be used to clean up handlers elsewhere as well. The alternative to the conditional in the function body would be to modify custom components (like PhoneInput) to pass the entire event object to the onChange prop instead of just a value.

nathanielparke:
Can you post a picture of the saved view?

m-j-mcdonald:
I didn't create a separate saved view, just so that the advisor could make multiple changes changes in a row without any hassle. The indication that it's saved is the save button shows it's loading.

kmeht:
Exciting! I'll review thoroughly later, but one comment is be wary of naming things. "modified" made me think it was a boolean at first, until I noticed it was a time. A better name is "modified_at", because it's very clear that it's a timestamp.

huxley/core/models.py
```diff
@@ -123,6 +124,8 @@ class School(models.Model):
     fees_owed = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('0.00'))
     fees_paid = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('0.00'))
 
+    modified = models.DateTimeField(default=datetime.utcnow)
```
kmeht: You'll want to use the `auto_now` and `auto_now_add` options instead: https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.DateField.auto_now

huxley/api/serializers/user.py
```diff
@@ -7,21 +7,39 @@
...
         fields = ('id', 'username', 'first_name', 'last_name', 'user_type',
                   'school', 'committee',)
-        read_only_fields = ('id', 'username', 'first_name', 'last_name',
-                            'user_type', 'committee',)
+        read_only_fields = ('id', 'username', 'user_type', 'committee',)
 
+    def update(self, instance, validated_data):
+        if validated_data.get('school'):
+            school_data = validated_data.pop('school')
+            School.objects.filter(name=instance.school).update(**school_data)
```
kmeht: Why is this call here? I'd imagine the implementation in the superclass should take care of it already, right?

huxley/api/serializers/user.py
```diff
@@ -7,21 +7,39 @@
...
+            User.objects.filter(username=instance).update(**validated_data)
```
kmeht: Same as above. Also, `instance` seems to suggest that it's an instance of a User... so why are we re-fetching the user? (Although that's a moot comment since I think we don't need to manually update).

huxley/api/serializers/user.py
```diff
@@ -7,21 +7,39 @@
...
+                      'New registraion information for {0}: \n\n'.format(instance.school) \
+                      + 'Advisor: \n' \
+                      + '\n'.join(['{0}: {1}'.format(field, validated_data[field]) for field in validated_data]) \
+                      + '\n\nSchool: \n' \
+                      + '\n'.join(['{0}: {1}'.format(field, school_data[field]) for field in school_data]),
+                      'tech@bmun.org',
+                      ['external@bmun.org'], fail_silently=False)
+            return instance
+        else:
+            return super(UserSerializer, self).update(instance, validated_data)
```
kmeht: Oh I see, we only want to update if the school has changed. Would the following work?
```python
def update(self, instance, validated_data):
    result = super(UserSerializer, self).update(instance, validated_data)
    if 'school' in validated_data:
      # send email...
    return result
    
```

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +324,105 @@ var AdvisorProfileView = React.createClass({
...
+    return null;
+  },
+
+  _handleChange: function(fieldName, event) {
+    var change = {};
+
+    // Handles both changes from HTMl and custom components
+    if (event.target) {
+      change[fieldName] = event.target.value;
```
kmeht: I'm pretty sure we're using Babel to compile our JS now, so we can use ES2015 computed properties:
```js
this.setState({
  [fieldName]: event.target ? event.target.value : event,
});
```

Give it a shot!

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +324,105 @@ var AdvisorProfileView = React.createClass({
...
+      contentType: 'application/json'
+    });
+    event.preventDefault();
+  },
+
+  _handleSuccess: function(data, status, jqXHR) {
+    this.setState({
+      errors: {},
+      loading: false
+    }, this.onSuccess);
```
kmeht: What's `onSuccess`? I don't see it defined anywhere...

huxley/www/static/js/huxley/components/AdvisorProfileView.js
```diff
@@ -243,9 +324,105 @@ var AdvisorProfileView = React.createClass({
...
+        <label className="hint error">
+          {this.state.errors.school[field]}
+        </label>
+      );
+    }
+
+    return null;
+  },
+
+  _handleChange: function(fieldName, event) {
```
tdowds: Might as well put this in its own file so we can import it into other components.

tdowds:
This is awesome, Michael. How do the errors render on the view?

m-j-mcdonald:
Errors screenshot:
screen shot 2016-07-18 at 9 54 43 pm

huxley/core/models.py
```diff
@@ -123,6 +124,8 @@ class School(models.Model):
     fees_owed = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('0.00'))
     fees_paid = models.DecimalField(max_digits=6, decimal_places=2, default=Decimal('0.00'))
 
+    modified_at = models.DateTimeField(default=timezone.now)
+
```
m-j-mcdonald: auto_now doesn't play nicely here because it needs to adjust when update() is called on schools, and the default update() method in django does not call save(), which is the trigger for auto_now

huxley/api/serializers/user.py
```diff
@@ -7,21 +7,39 @@
...
-        read_only_fields = ('id', 'username', 'first_name', 'last_name',
-                            'user_type', 'committee',)
+        read_only_fields = ('id', 'username', 'user_type', 'committee',)
 
+    def update(self, instance, validated_data):
+        if 'school' in validated_data:
+            school_data = validated_data.pop('school')
+            school_data['modified_at'] = timezone.now()
+            School.objects.filter(name=instance.school).update(**school_data)
+            send_mail('{0} has updated its registration information'.format(instance.school),
```
m-j-mcdonald: Update needs to be called explicitly on the school here because django's default update method doesn't handle nested serializers

huxley/api/serializers/user.py
```diff
@@ -7,21 +7,39 @@
...
                   'school', 'committee',)
-        read_only_fields = ('id', 'username', 'first_name', 'last_name',
-                            'user_type', 'committee',)
+        read_only_fields = ('id', 'username', 'user_type', 'committee',)
 
+    def update(self, instance, validated_data):
+        if 'school' in validated_data:
+            school_data = validated_data.pop('school')
+            school_data['modified_at'] = timezone.now()
+            School.objects.filter(name=instance.school).update(**school_data)
```
kmeht: It looks like we're relying on the string representation of a school to be its name (which is reasonable, but subject to change). Instead, we should look it up by ID. Does the following work?
```python
School.objects.filter(id=instance.school.id).update(**school_data)
```

kmeht:
LGTM, one last comment on my end. @tdowds, @nathanielparke, anything else?

tdowds:
Assuming the app still works, this looks really great to me. We'll hopefully get some more feedback from external and the advisors as they use it. If there's anything we need to improve on, I'll let you you guys know

nathanielparke:
This looks good to me as well. Great job @m-j-mcdonald!  I'm going to merge.

Issue #472, opened 2016-06-18
Introduce autoformatters into the workflow


kmeht:
There are automatic code format tools that should help us avoid wasting code review cycles on style nits. One example is [`autopep8`](https://pypi.python.org/pypi/autopep8) for Python.

I believe it has an option to specify line numbers, so we could write a Fabric command to detect the changed lines in a Python file, and run the autoformatter on it. This would help us move faster in development and code review.

nathanielparke:
Another possible one is https://github.com/google/yapf

PR #496, opened 2016-08-11
Introduce python autoformatter

m-j-mcdonald:
This adds a python auto formatter as discussed in #472. I introduced a command 'fab format' to run the auto formatter, which also gets run as part of submit. As an example of what the auto formatter might do, look at how the code of __init__.py and git.py was modified.

### DIFF
--- a/fabfile/__init__.py
+++ b/fabfile/__init__.py
@@ -45,10 +45,11 @@ def format():
     py_diff_list = [pyfile for pyfile in diff_list if pyfile.endswith('.py')]
 
     if confirm('Review formatting changes? (Select no to approve all)'):
+
         for pyfile in py_diff_list:
             print('\n')
             for line in FormatFile(pyfile, print_diff=True):
-                print(line)
+                print(line+'*')
             if confirm('Accept changes to %s?' % pyfile):
                 FormatFile(pyfile, in_place=True)
     else:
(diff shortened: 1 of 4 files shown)


fabfile/__init__.py
```diff
@@ -22,7 +23,8 @@ def feature(branch_name=None):
         dependencies.check()
         migrations.check()
     else:
-        print ui.error('No branch name given. Usage: fab feature:<branch_name>')
+        print ui.error(
+            'No branch name given. Usage: fab feature:<branch_name>')
```
m-j-mcdonald: This is an example of the formatting

fabfile/__init__.py
```diff
@@ -37,8 +39,32 @@ def update():
 
 
 @task
+def format():
+    '''Format and commit python files committed on the current feature branch'''
+    diff_list = git.diff_name_only()
+    py_diff_list = [pyfile for pyfile in diff_list if pyfile[-3:] == '.py']
```
kmeht: `if pyfile.endswith('.py')`

fabfile/__init__.py
```diff
@@ -37,8 +39,32 @@ def update():
...
+            if confirm('Accept changes to %s?' % pyfile):
+                FormatFile(pyfile, in_place=True)
+    else:
+        for pyfile in py_diff_list:
+            FormatFile(pyfile, in_place=True)
+    print ui.info('Formatting complete')
+
+    if len(local('git status --porcelain', capture=True)) > 0:
+        local('git add --all')
+        local('git commit -m "Ran autoformatter"')
```
kmeht: These two lines can be collapsed into `git commit -am "Ran autoformatter"`

fabfile/__init__.py
```diff
@@ -37,8 +39,32 @@ def update():
...
+
+
+@task
 def submit(remote='origin', skip_tests=False):
     '''Push the current feature branch and create/update pull request.'''
+    execute(format)
```
kmeht: What's going on here? We can't just call the function?

kmeht:
Great diff! Does yapf seem to be the standard formatter these days? Does it do other stuff than whitespace? Do we want to include options for other formatting as well?

m-j-mcdonald:
Autopep8 seems to still be more common but people prefer yapf. I believe it only handles whitespace, but the advantage is that instead of just fixing pep8 violations it reformats the whole file to follow a consistent style.

kmeht:
Ok, let's see how this formatter works for us for a while. Wouldn't be too hard to swap out the formatter later if we so choose.

PR #497, opened 2016-08-16
Switch to Webpack for JS.

### LINKED ISSUE #387: Investigate Webpack
WebPack is an alternative to Browserify. Let's investigate it and determine whether it'd be more appropriate for us to use.

kmeht:
Ditches pipeline+browserify for JS. Removes the browserify dependencies, pipeline settings, and adjusts the deploy command in fabric to build the JS before running collectstatic.

Resolves #387.

### DIFF
--- a/huxley/www/templates/www.html
+++ b/huxley/www/templates/www.html
@@ -19,7 +19,7 @@
             ContactTypes = {{ contact_types|safe }};
             ProgramTypes = {{ program_types|safe }};
         </script>
-        {% javascript 'huxley' %}
+        <script type="text/javascript" src="/static/js/bundle.js"></script>
     {% endblock js %}
 </head>
 <body>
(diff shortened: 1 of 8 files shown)


huxley/www/templates/www.html
```diff
@@ -19,7 +19,7 @@
             ContactTypes = {{ contact_types|safe }};
             ProgramTypes = {{ program_types|safe }};
         </script>
-        {% javascript 'huxley' %}
+        <script type="text/javascript" src="/static/js/bundle.js"></script>
```
tdowds: If this is an absolute path what is the root directory here? `www`? Just want to make sure this path points to `./huxley/www/static/js` as it does [here] (https://github.com/bmun/huxley/commit/2da93a3b950e9c53ab79bb60e6312cff5ce58609#diff-11e9f7f953edc64ba14b0cc350ae7b9dR22)

kmeht: It's not an absolute path; in HTML it's a relative URL. So, if the domain were stage.huxley.bmun.org, it would resolve to stage.huxley.bmun.org/static/js/bundle.js. The /static/ path maps to the static applications on the webserver, which will serve the static file.

PR #517, opened 2016-09-05
npm script to build and start server

tdowds:
Added two commands to
1) Build our code so we don't have to use `./node_modules/.bin/webpack --watch` to build
2) Build our code and start our server

Usage:
1) `npm run build`
2) `npm start` or `npm run start`


tdowds:
Updated so that `npm start` will build and then run the server. It works, however this build can't watch for changes and start the server simultaneously. So after you run the command, the code will build and the server will run, however if you make any changes on the front end they won't be rebuilt automatically.

package.json
```diff
@@ -17,7 +17,9 @@
     "react-router": "1.0.0"
   },
   "scripts": {
-    "test": "jest"
+    "test": "jest",
+    "build": "./node_modules/.bin/webpack --progress --watch",
+    "start": "./node_modules/.bin/webpack; python manage.py runserver"
```
kmeht: How about this: we get rid of the start command, and developers can `npm build` and `python manage.py runserver` in separate tabs. It would be cool to do it both in one command, but this would be a good start for now.

Issue #294, opened 2014-09-02
Integrate a Python linter into the workflow.


kmeht:
It would be great to catch small style issues, as well as other issues that are better caught via static analysis. We should add a linter that runs before submit, so we can fix these issues before submitting PRs.

kmeht:
@m-j-mcdonald, this one might be up your alley, since you added the Python formatter also.

PR #528, opened 2016-10-09
Add pylint to workflow

m-j-mcdonald:
Sorry about the initial long title, I forgot where I was entering the stuff. 

This adds Pylint to the workflow. Upon submit users will have the option to view a brief Pylint evaluation of the python files that were changed. They can then choose to continue with the submit based off the evaluation. This only prints an evluation, it does not alter the code in anyway (which is why I seperated it from the format command). This address issue #294.

### DIFF
--- a/fabfile/__init__.py
+++ b/fabfile/__init__.py
@@ -42,7 +42,7 @@ def update():
 
 @task
 def lint():
-    '''Format and commit python files committed on the current feature branch'''
+    '''Run linter evaluation of files committed on current feature branch'''
     diff_list = git.diff_name_only()
     py_diff_list = [pyfile for pyfile in diff_list if pyfile.endswith('.py')]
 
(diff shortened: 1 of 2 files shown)


m-j-mcdonald:
Also, I did not go into any customization yet so this just runs Pylint as is

kmeht:
LGTM! Could you include a paste of a sample lint output? Curious as to how noisy lint is without any customization.

fabfile/__init__.py
```diff
@@ -62,6 +77,11 @@ def format():
 @task
 def submit(remote='origin', skip_tests=False):
     '''Push the current feature branch and create/update pull request.'''
+    if confirm('View linter evluation? (Will not change code)'):
```
kmeht: "evaluation"

m-j-mcdonald:
screen shot 2016-10-09 at 3 19 18 pm
This is an example with some random mistakes.

kmeht:
Weird, only the first one seems like a legit error. The rest are just plain wrong. Maybe we have to configure it to understand Python 2.7?

m-j-mcdonald:
I introduced those errors deliberately to show the functionality

kmeht:
Oh, rad. This is good to go, then.

Issue #519, opened 2016-09-06
Fix text input on refocus


m-j-mcdonald:
Right now there's an issue where every time a character is entered into an input box the focus shifts to the end of the line immediately after registering the keystroke. So if someone tries to go back and edit part of what they were typing, only a single character will go where they want it and the rest of the keystrokes will be registered at the end of the text. E.g., if someone was typing "2700 Avenue" and realized it should have Durant in there, it would come out as "2700 DAvenueurant " when they tried to fix it.

kmeht:
Bunch of discussion here, maybe it'll be useful: https://github.com/facebook/react/issues/955

PR #530, opened 2016-10-16
Fix text refocus per #519

m-j-mcdonald:
This is the simplest resolution to the issue with text refocus discussed in #519. By leaving text inputs as uncontrolled components, the user's input is still preserved internally but the JS no longer resets the value every render. What was happening was the user would type, React would handle the change & update the state, then it would re-render and in doing so look at the value prop for that input (which was set to retrieve its data from the state), obliterate what the user had typed and replace it with the state (which was the same as what the user had typed). This meant cursor position was not being preserved. This just removes the explicit setting of the value prop (which doesn't change how the state is updated, just how the user sees it)


m-j-mcdonald:
I have no idea what's going on with the static stuff in this diff. I'll clean that up

kmeht:
This diff will break registration. On submit, we send all the values from the state to the server, but by making these uncontrolled components, we're not updating the state at all.

m-j-mcdonald:
I tested that. The state is updated still, it just isn't reflected in how the text inputs are rendered (which shouldn't matter, since the state should mirror what the user typed)

kmeht:
Ok, that's reasonable.

For inputs that already have text in them (like AdvisorProfileView), use the `defaultValue` prop instead of the placeholder to show it. That's only used on initial render for uncontrolled components.

Also, can we document this hack somewhere? Using uncontrolled inputs everywhere opens us up to bugs if we ever have features that have to update inputs as a result of anything other than users typing.

kmeht:
@m-j-mcdonald another thing we can do to minimize the scope of this change is to actually modify the `TextInput` component to not pass the value into the input component. That way only that component needs to be changed, nothing else.

It can also take an `isControlled` prop that overrides the behavior to make any particular instance a controlled component if we ever need to.

m-j-mcdonald:
Okay, I really don't get why the static files keep getting added but I will remove them

m-j-mcdonald:
This works well on my end. Each component has the expected behavior

huxley/www/js/components/TextInput.js
```diff
@@ -11,21 +11,36 @@ var cx = require('classnames');
 var TextInput = React.createClass({
```
kmeht: Let's add a docblock to this component describing why we're making them uncontrolled by-default, and linking to a task number.
```js
/**
 * TextInput is uncontrolled by default because blah blah...
 * <links to relevant task>
 */
var TextInput = React.createClass({...});
```

kmeht:
Looks great to me! Two comments on the TextInput component, and this is good to go.

PR #533, opened 2016-10-23
Added notifications

### LINKED ISSUE #495: Indicate to advisors a successful save or error
When an advisor successfully saves any of their information, we don't explicitly tell the advisor their save was successful. This includes hitting save on the AdvisorProfileView and on the AdvisorAssignmentView when delegates are successfully assigned to assignments. We should have some kind of banner or UI element that informs the advisor that the save was successful (I'm thinking some kind of banner that drops down from the top of the screen, but really anything that's green and says "Success!" should do).

michaeleliot:
Added [react-notification-system](https://github.com/igorprado/react-notification-system) to display a notification when an advisor saves there information. Closes #495


michaeleliot:
screen shot 2016-10-23 at 5 24 26 pm

Isn't www/js/components the top level component? If not, then where would that be?

tdowds:
Also when you're committing, do a `git status` before doing `git add` so you know exactly which files you're adding. There's no reason to be committing `huxley/www/static/js/bundle.js`

tdowds:
went with a534793084056d02a5226538cf8cc2859562f38c instead

Issue #523, opened 2016-10-02
Use Flux for data updates


kmeht:
In many places, such as AdvisorAssignmentsView, we are updating the component's state directly and then "refreshing" by navigating to the same page to update.

Instead, we should:
- subscribe to the Stores in the component
- update the data in the Stores via Actions
- then, the component will re-render automatically in response to the changed data.

This will require adding Actions for each type of data update, and updating the Stores to handle such actions. Admittedly, this is a bunch of boilerplate for a small app. But, it will help the app scale as we add more and more features.

tdowds:
Things to do:
- [ ] Use flux for Delegates
- [ ] Use flux for Assignments
- [ ] Fire actions in response to successful saves or errors from the server
- [ ] Use flux to get committees and countries(?)

Issue #518, opened 2016-09-05
Deleting delegates doesn't update client store


kmeht:
When we delete a delegate, the clientside store isn't updated.

Repro steps:
- On roster page, delete a delegate
- Switch to another page and back
- See that the delegate is still there (note that on refresh, the delegate is gone).

This is because we're not following Flux well. In the roster view, we should be dispatching an action that both sends the server API request and also removes the delegate from the clientside store.

kmeht:
@tdowds, this is one specific case of #523.

tdowds:
Closed via 408addb76ca7393c771af666c2a4a7d85dfed508

PR #535, opened 2016-11-03
uses flux for deleting delegates.

tdowds:
Using this PR to use flux for data updates (#523).

For this particular commit it handles deleting a delegate (#518). I'd like to get a review on this before implementing other updates. Furthermore I should probably update the DelegateStore test to reflect this change.


huxley/www/js/actions/DelegateActions.js
```diff
@@ -0,0 +1,37 @@
...
+'use strict';
+
+var ActionConstants = require('constants/ActionConstants');
+var Dispatcher = require('dispatcher/Dispatcher');
+
+var DelegateActions = {
+  deleteDelegate(delegate) {
+    Dispatcher.dispatch({
+      actionType: ActionConstants.DELETE_DELEGATE,
+      delegate: delegate
```
kmeht: We should only need a `delegateID` to delete a delegate, not the entire object.

huxley/www/js/actions/DelegateActions.js
```diff
@@ -0,0 +1,37 @@
...
+      actionType: ActionConstants.ADD_DELEGATE,
+      delegate: delegate
+    });
+  },
+
+  updateDelegate(delegateID, name, email, schoolID) {
+    Dispatcher.dispatch({
+      actionType: ActionConstants.UPDATE_DELEGATE,
+      delegateID: delegateID,
+      name: name,
```
kmeht: Why are all the fields spread out here, vs. the single object in the add method? We should be consistent; I vote for the object here.

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -57,6 +58,18 @@ var AdvisorRosterView = React.createClass({
...
+  componentDidMount: function() {
+    DelegateStore.addListener(function() {
+      var schoolID =  CurrentUserStore.getCurrentUser().school.id;
+      var delegates = DelegateStore.getDelegates(schoolID);
+      this.setState({
+        delegates: delegates,
+        loading: false,
+        modal_open: false
+      });
+    }.bind(this));
```
kmeht: Use an arrow function so we don't have to explicitly bind.
```js
DelegateStore.addListener(() => {
  //...
});
```

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -57,6 +58,18 @@ var AdvisorRosterView = React.createClass({
     Modal.setAppElement('body')
   },
 
+  componentDidMount: function() {
+    DelegateStore.addListener(function() {
+      var schoolID =  CurrentUserStore.getCurrentUser().school.id;
+      var delegates = DelegateStore.getDelegates(schoolID);
+      this.setState({
+        delegates: delegates,
+        loading: false,
```
kmeht: Don't think we need to set either loading or modal_open in this callback. What was the reason for including them?

huxley/www/js/stores/__tests__/DelegateStore-test.js
```diff
@@ -8,18 +8,20 @@
 jest.dontMock('stores/DelegateStore');
 
 describe('DelegateStore', () => {
+  var ActionConstants
```
kmeht: semicolon

huxley/www/js/stores/__tests__/DelegateStore-test.js
```diff
@@ -40,4 +42,41 @@ describe('DelegateStore', () => {
       }),
     ]);
   });
+
+  it('adds a delegate', () => {
+    return Promise.all([
```
kmeht: Don't need `Promise.all` if there's only one promise

huxley/www/js/stores/__tests__/DelegateStore-test.js
```diff
@@ -40,4 +42,41 @@ describe('DelegateStore', () => {
...
+        expect(DelegateStore.getDelegates(mockSchoolId).length).toEqual(1);
+      })
+    ]);
+  });
+
+  it('updates a delegate', () => {
+    return Promise.all([
+      DelegateStore.getDelegates(mockSchoolId, (delegates) => {
+        expect(delegates).toEqual(mockDelegates);
+        var updatedDelegate = mockDelegates[0]
```
kmeht: I don't think this test works as expected, because `delegates` and `mockDelegates` are literally the same object. That means, if you update one delegate's name in-place (like below), it already updates it in the store (which is no bueno).

Instead, just write out the new name in the `updateDelegate` call:
```js
DelegateStore.updateDelegate(
  d.id,
  'Jake Moskowitz',
  d.email,
  mockSchoolId,
);
```

huxley/www/js/actions/DelegateActions.js
```diff
@@ -0,0 +1,37 @@
...
+      email: email,
+      schoolID: schoolID
```
kmeht: I don't think schoolID should be an updateable field. Opens us up to bugs where delegates are in the wrong school. we should also make this immutable on the server side, if we haven't already.

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,79 @@
...
+        callback(value);
+      });
     }
-    return _delegatePromises[schoolID];
+    return _delegates;
+  }
+
+  deleteDelegate(delegate) {
+    ServerAPI.deleteDelegate(delegate.id);
+    var index = _delegates.indexOf(delegate);
```
kmeht: Assuming that we'd only pass the `delegateID` into this function:
```js
_delegates = _delegates.filter(d => d.id !== delegateID);
```

This sets `_delegates` to be a new array instead of mutating the old one. In general, we want to avoid mutative behaviors like this, because we don't know who else has a reference to these objects.

tdowds:
Updated this PR to do the following:
1. Does not depend on Promise in the Delegate Store
2. Updated Delegate Store methods to reflect changes in (1)
3. Updated AdvisorAssignmentView to reflect changes in 1 and 2 (Added a bulk update action on delegates for this)
4. Updated Delegate Store tests accordingly

Things still to do:
1. Use flux in AssignmentStore
2. Fire actions on successful or unsuccessful ServerAPI requests and update UI accordingly
For these two items I think I will open in separate diffs after this one is good to go

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -50,13 +51,23 @@ var AdvisorRosterView = React.createClass({
       )});
     }.bind(this));
 
-    DelegateStore.getDelegates(user.school.id, function(delegates) {
-      this.setState({delegates: delegates});
-    }.bind(this));
+    this.setState({delegates: DelegateStore.getDelegates(user.school.id)});
```
kmeht: This now belongs in `getInitialState`

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,94 @@
...
+    _schoolsDelegates[delegate.school].push(delegate);
+  }
+
+  updateDelegate(delegateID, delta) {
+    ServerAPI.updateDelegate(delegateID, {
+      name: delta.name,
+      email: delta.email,
+    });
+    var delegate = _delegates[delegateID];
+    delegate.name = delta.name;
```
kmeht: This is still a mutative action. Instead, I'd use the spread operator to copy the delegate:
```js
const delegate  = {
  ..._delegates[delegateID],
  name,
  email,
};
_delegates[delegateID] = delegate;
_schoolsDelegates[delegate.school] = /*...*/;
```

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,94 @@
...
+  deleteDelegate(delegateID) {
+    ServerAPI.deleteDelegate(delegateID);
+    var schoolID = _delegates[delegateID].school;
+    delete _delegates[delegateID];
+    _schoolsDelegates[schoolID] = _schoolsDelegates[schoolID].filter(d => d.id !== delegateID);
+  }
+
+  addDelegate(delegate) {
+    _delegates[delegate.id] = delegate;
+    _schoolsDelegates[delegate.school].push(delegate);
```
kmeht: This is a mutation, I'd use `concat` instead

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,94 @@
...
+  updateDelegate(delegateID, delta) {
+    ServerAPI.updateDelegate(delegateID, {
+      name: delta.name,
+      email: delta.email,
+    });
+    var delegate = _delegates[delegateID];
+    delegate.name = delta.name;
+    delegate.email = delta.email;
+    _delegates[delegateID] = delegate;
+    _schoolsDelegates[delegate.school] = _schoolsDelegates[delegate.school].map(d => d.id == delegate.id ? delegate : d);
```
kmeht: Same style nit, put the function on a separate line

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,94 @@
...
+  }
+
+  updateDelegates(schoolID, delegates) {
+    ServerAPI.updateSchoolDelegates(
+      schoolID,
+      JSON.stringify(delegates)
+    )
+    for (var i = 0; i < delegates.length; i++) {
```
kmeht: Use a `for-of` loop instead:
```js
  //...
}
```

huxley/www/js/stores/DelegateStore.js
```diff
@@ -1,32 +1,94 @@
...
+  getDelegates(schoolID) {
+    if (_schoolsDelegates[schoolID]) {
+      return _schoolsDelegates[schoolID];
     }
-    if (callback) {
-      _delegatePromises[schoolID].then(callback);
+
+    ServerAPI.getDelegates(schoolID).then(value => {
+      _schoolsDelegates[schoolID] = value;
+      for (var i = 0; i < value.length; i++) {
```
kmeht: Use a `for-of` loop.

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -63,7 +64,18 @@ var AdvisorAssignmentsView = React.createClass({
       }
       this.setState({countries: new_countries})
     }.bind(this));
-    DelegateStore.getDelegates(user.school.id, (delegates) => {
+    var delegates = DelegateStore.getDelegates(user.school.id);
```
kmeht: Put the following in `getInitialState`

PR #538, opened 2016-11-10
Added committee preferences column when downloading assignment preferences from admin portal

tdowds:
External expressed a need to include the committee preferences of a school along with a school's country preferences when downloading those preferences from the admin panel. This diff

1. Adds a column that lists a school's committee preferences if any
2. Updates tests to reflect that
3. updates the .gitignore file to ignore `huxley/www/static/js/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,4 @@ htmlcov/
 
 node_modules/
 
-huxley/wwww/static/js/
+huxley/www/static/js/
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ fabfile/deploy.py
 htmlcov/
 
 node_modules/
+
+huxley/wwww/static/js/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -21,3 +21,5 @@ fabfile/deploy.py
 htmlcov/
 
 node_modules/
+
+huxley/www/static/js/
```
kmeht: Make this more specific, just in case we throw more JS files in there that we actually want to track.

huxley/core/admin/schools.py
```diff
@@ -57,99 +58,88 @@ def info(self, request):
...
 
         return schools
 
     def preferences(self, request):
         ''' Returns a CSV file containing the current set of
             Schools registered with all of its fields. '''
         schools = HttpResponse(content_type='text/csv')
-        schools['Content-Disposition'] = 'attachment; filename="preferences.csv"'
+        schools[
+            'Content-Disposition'] = 'attachment; filename="preferences.csv"'
```
kmeht: really? this is what the autoformatter picks? So be it.

PR #539, opened 2016-11-15
important fixes

tdowds:
This

1. Fixes the memory leak in our components subscribing to DelegateStore
2. Removes white space when loading assignments
3. Adds a forgotten break statement in the dispatch method of DelegateStore
4. Removes unneeded Assignment code in AdvisorRosterView

### DIFF
--- a/huxley/www/js/components/AdvisorAssignmentsView.js
+++ b/huxley/www/js/components/AdvisorAssignmentsView.js
@@ -82,7 +82,7 @@ var AdvisorAssignmentsView = React.createClass({
   },
 
   componentWillUnmount: function() {
-    this._delegatesToken.remove();
+    this._delegatesToken && this._delegatesToken.remove();
   },
 
   render: function() {
--- a/huxley/www/js/components/AdvisorAssignmentsView.js
+++ b/huxley/www/js/components/AdvisorAssignmentsView.js
@@ -40,7 +40,8 @@ var AdvisorAssignmentsView = React.createClass({
       committees: {},
       countries: {},
       delegates: delegates,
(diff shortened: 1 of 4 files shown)


huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -81,6 +81,10 @@ var AdvisorAssignmentsView = React.createClass({
     });
   },
 
+  componentWillUnmount: function() {
+    this._delegatesToken.remove();
```
kmeht: Guard against the possibility of the component never mounting with:
```js
this._delegatesToken && this._delegatesToken.remove();
```

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -68,6 +56,10 @@ var AdvisorRosterView = React.createClass({
     });
   },
 
+  componentWillUnmount: function() {
+    this._delegatesToken.remove();
```
kmeht: ditto

huxley/www/js/stores/DelegateStore.js
```diff
@@ -78,6 +78,7 @@ class DelegateStore extends Store {
         break;
       case ActionConstants.UPDATE_DELEGATES:
         this.updateDelegates(action.schoolID, action.delegates);
+        break;
```
kmeht: Didn't we have a test case that would've caught this?

tdowds: I think the test tests whether the store was updated, not whether the store was updated and a change emitted

Issue #507, opened 2016-08-28
Get and Update delegates based off committee


tdowds:
In order to get all the delegates in a committee loaded onto a page in the chair app and to update those delegates we need to be able to filter delegates by committee. To do this we should create a url and view on the Delegate side to filter delegate's by the committee field of its assignment field. We would get delegates by a call in the delegate store. We would update delegates by making an ajax request to that url.

Issue #504, opened 2016-08-28
ChairAttendanceView component


tdowds:
Create a view to allow chairs to track the attendance of delegates. Attendance should be based off delegation (not delegate). We should have one radio select per session. Attendance should be updated in bulk.

tdowds:
closed via f006a47670353f24d9e24fe9f80bc464a4cf28be

PR #542, opened 2016-12-04
ChairAttendanceView frontend

m-j-mcdonald:
This is the code for the frontend of tracking delegate attendance. We have tested it with corresponding changes to the backend (which will be in a separate pull request). This is mainly issue #504, and part of #507. @michaeleliot @jmosky12


huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -32,6 +37,31 @@ var ChairAttendanceView = React.createClass({
...
+    CountryStore.getCountries(function(countries) {
+      countries = countries.filter(country => 
+        country.id in country_assignments
+      );
+      for (var country of countries) {
+        country_assignments[country.name] = country_assignments[country.id];
+        delete country_assignments[country.id];
+      }
+      this.setState({country_assignments: country_assignments});
+    }.bind(this));
```
tdowds: `countries =>` so you don't have to explicitly bind `this` here

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -47,9 +77,10 @@ var ChairAttendanceView = React.createClass({
               <thead>
                 <tr>
                   <th>Assignment</th>
-                  <th>Present</th>
-                  <th>Present2</th>
-                  <th>Present3</th>
+                  <th>Friday</th>
```
tdowds: As I mentioned in the other PR, Session One, Session Two, Session Three, and Session Four might be more manageable names

huxley/www/js/lib/ServerAPI.js
```diff
@@ -34,6 +34,20 @@ var ServerAPI = {
   },
 
   /**
+   * Get a list of all assignments for the given school ID.
```
kmeht: committee ID

huxley/www/js/lib/ServerAPI.js
```diff
@@ -34,6 +34,20 @@ var ServerAPI = {
...
+   */
+  getCommitteeAssignments(committeeID) {
+    return _get(`/api/assignments/committee/${committeeID}`);
```
kmeht: IMO we shouldn't make another API endpoint for this. It's really just a filter on the assignments list API. We should then customize the `get` behavior of the AssignmentListView to modify the queryset if any filters are present.

The URI would then become something like `/api/assignments?committee_id=12`.

kmeht: (I haven't looked at the backend PR yet, but I'm just taking a guess based on the URI that it'll be a new API endpoint.

huxley/www/js/lib/ServerAPI.js
```diff
@@ -34,6 +34,20 @@ var ServerAPI = {
...
+  },
+
+  /**
+   * Get a list of all delegates for the given committee ID.
+   */
+  getCommitteeDelegates(committeeID) {
+    return _get(`/api/delegates/committee/${committeeID}`);
```
kmeht: Very similar comment here.

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -67,48 +99,107 @@ var ChairAttendanceView = React.createClass({
...
-    //       </td>
-    //     </tr>
-    //   )}.bind(this));
-    return (
-      <tr></tr>
-    );
+    var countries = Object.keys(this.state.country_assignments);
+    return countries.map(country => {
+      return (
+        <tr>
```
kmeht: This attendance row is pretty huge. Let's extract it into its own component.

huxley/www/js/stores/DelegateStore.js
```diff
@@ -13,6 +13,7 @@ var {Store} = require('flux/utils');
 
 
 var _schoolsDelegates = {};
+var _committeeDelegates = {};
```
kmeht: This data is getting really complicated. What if we someday want to query delegates by country? Will we have a fourth map? A fifth?

We should figure out a way to query delegates by school and committee, instead of having three different ways of fetching and storing them. School should be pretty easy, since it's a field on the delegate object itself. e.g, `getDelegatesForSchool` or something could be a simple filter statement (if we choose to store them as an array):

```js
return _delegates.filter(d => d.school_id === schoolID);
```

Committees will be a little harder, since we have to go through the assignment store. Any ideas on how we could model this?

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -67,48 +99,107 @@ var ChairAttendanceView = React.createClass({
...
+      delegate[session] = !delegate[session];
+    }
+    country_assignments[country] = delegates;
+    this.setState({
+      country_assignments: country_assignments
+    });
+  }, 
+
+  _handleSaveAttendance(event) {
+    this.setState({loading: true});
```
tdowds: Since we're updating optimistically, I don't think we need to show the loading wheel

m-j-mcdonald:
The new commits incorporate the recommended changes but also take a different approach to Assignment & Delegate store.

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -67,50 +100,59 @@ var ChairAttendanceView = React.createClass({
...
+
+  _handleSaveAttendance(event) {
+    var committee = CurrentUserStore.getCurrentUser().committee;
+    var country_assignments = this.state.country_assignments;
+    var delegates = [];
+    for (var country in country_assignments) {
+      delegates = delegates.concat(country_assignments[country]);
+    }
+    DelegateActions.updateCommitteeDelegates(committee, delegates);
+    window.alert("Attendance Saved.");
```
tdowds: Probably not worth throwing this up in case an error is thrown by the server. We can handle telling advisors their save was successful in a separate diff.

m-j-mcdonald: I could switch it to use .then and have a _handleSuccess and _handleError pretty easily, if that would be desirable

tdowds: a `.then` on `DelegateActions.updateCommitteeDelegates`? Does that return a Promise?

huxley/www/js/stores/DelegateStore.js
```diff
@@ -6,54 +6,70 @@
...
   updateDelegates(schoolID, delegates) {
     ServerAPI.updateSchoolDelegates(schoolID, delegates);
     for (const delegate of delegates) {
       _delegates[delegate.id] = delegate;
     }
-    _schoolsDelegates[schoolID] = delegates;
+  }
+
+  updateCommitteeDelegates(committeeID, delegates) {
+    ServerAPI.updateCommitteeDelegates(committeeID, delegates)
```
tdowds: semicolon

huxley/www/js/components/ChairAttendanceView.js
tdowds: bump on this

tdowds:
This is some awesome work! A couple things:
1. Change "Friday Night", "Saturday Morning" ... to "Session One", "Session Two"... on the view itself
2. We need to update AdvisorRosterView with disabled checkboxes so they can see their delegates' attendance.

huxley/www/js/components/ChairAttendanceView.js
m-j-mcdonald: Oops, I was thinking of the wrong thing. Yeah, I'll take this out

m-j-mcdonald:
The second part we're good for, we were going to add something along the line of an AdvisorFeedackView to show attendance & summaries, just to keep the roster view uncrowded & allow better formatting for the feedback stuff. That'll probably be in a pr later this week. SummaryView's good to go and I'll probably post the pr once this one's closed.

tdowds:
This looks pretty good to me! What do you think @kmeht?

Issue #506, opened 2016-08-28
Update Delegate model to track attendance


tdowds:
In order to include a feature for delegate attendance in the chair app, we need to store that information in our delegate model. Therefore we need to add fields for each session attendance. For now these can be booleans but we should build in anticipation for marking "Present", "Present and Voting", and "Not Present".

PR #543, opened 2016-12-04
Committee attendance backend

m-j-mcdonald:
Here is the backend for tracking committee attendance. This covers issue #506 and most of #507. @michaeleliot @jmosky12


m-j-mcdonald:
Note: the auto formatter made some weird changes.

huxley/api/permissions.py
```diff
@@ -28,6 +28,14 @@ def has_object_permission(self, request, view, obj):
         return request.user.is_superuser or request.user == obj.advisor
 
 
+class IsChairOrSuperuser(permissions.BasePermission):
+    '''Accept only the school's advisor or superusers.'''
```
tdowds: chair, not advisor

huxley/api/permissions.py
```diff
@@ -28,6 +28,14 @@ def has_object_permission(self, request, view, obj):
...
+
+    def has_object_permission(self, request, view, obj):
+        print(obj)
```
tdowds: remove this print statement

huxley/api/views/assignment.py
```diff
@@ -5,23 +5,38 @@
...
+
+
+class AssignmentCommitteeDetail(generics.ListAPIView):
+    authentication_classes = (SessionAuthentication, )
+    queryset = Assignment.objects.all()
+    serializer_class = AssignmentSerializer
+    permission_classes = (IsChairOrSuperuser, )
+
+    def get_queryset(self):
+        '''Filter schools by the given pk param.'''
```
tdowds: assignments, not schools

huxley/core/models.py
```diff
@@ -394,6 +394,11 @@ class Delegate(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     summary = models.TextField(default='', blank=True, null=True)
 
+    friday_attendance = models.BooleanField(default=False)
```
tdowds: a more manageable naming scheme might be "session_one, session_two, session_three, and session_four". They are technically committee sessions.

tdowds:
Would anyone be opposed to disabling the autoformatter for a little bit? Would make it easier to review in my opinion.

huxley/api/views/delegate.py
```diff
@@ -5,20 +5,38 @@
...
     serializer_class = DelegateSerializer
+
+
+class DelegateCommitteeDetail(generics.ListAPIView, ListUpdateModelMixin):
+    authentication_classes = (SessionAuthentication, )
+    queryset = Delegate.objects.all()
+    serializer_class = DelegateSerializer
+    permission_classes = (IsChairOrSuperuser, )
+
+    def get_queryset(self):
```
kmeht: As mentioned in the other PR, we can actually just override this method in the DelegateList view to either return `Delegate.objects.all()`, or filtered by the committee ID if that parameter is provided. Then, the URI would become `/api/delegates?committee_id=...`

huxley/api/views/assignment.py
```diff
@@ -5,23 +5,38 @@
...
         return self.partial_update(request, *args, **kwargs)
+
+
+class AssignmentCommitteeDetail(generics.ListAPIView):
+    authentication_classes = (SessionAuthentication, )
+    queryset = Assignment.objects.all()
+    serializer_class = AssignmentSerializer
+    permission_classes = (IsChairOrSuperuser, )
+
+    def get_queryset(self):
```
kmeht: Same comment as in DelegateCommitteeDetail.

huxley/api/views/delegate.py
```diff
@@ -5,20 +5,38 @@
...
+    authentication_classes = (SessionAuthentication, )
     queryset = Delegate.objects.all()
-    permission_classes = (IsSchoolDelegateAdvisorOrSuperuser,)
+    permission_classes = (IsSchoolDelegateAdvisorOrSuperuser, )
     serializer_class = DelegateSerializer
+
+
+class DelegateCommitteeDetail(generics.ListAPIView, ListUpdateModelMixin):
+    authentication_classes = (SessionAuthentication, )
+    queryset = Delegate.objects.all()
```
kmeht: I don't think you need this if you define a `get_queryset` method.

huxley/core/models.py
kmeht: Is attendance per-delegate, or per assignment? Like, when the chair calls out a country and only one delegate responds, how do you know who it is?

huxley/api/tests/school/test_school_delegates.py
```diff
@@ -109,43 +114,35 @@ def setUp(self):
...
             school=self.school,
-            assignment=self.assignment2
-        )
+            assignment=self.assignment2)
 
         self.params = [
-            {'id': self.delegate1.id, 'assignment': self.new_assignment.id},
-            {'id': self.delegate2.id, 'assignment': None}
+            {'id': self.delegate1.id,
+             'assignment': self.new_assignment.id}, {'id': self.delegate2.id,
```
kmeht: barf. perhaps we should investigate a new autoformatter.

huxley/core/models.py
tdowds: It's per assignment. We are (at least for now) just planning on duplicating the information we get per assignment to each delegate assigned to that assignment

m-j-mcdonald:
This incorporates the suggested changes, as well as introduces changes to the permissions and unit tests that were absent from the earlier pull request. The big lingering question is about what permissions a chair should have.

huxley/api/views/delegate.py
```diff
@@ -24,11 +24,18 @@ def get_queryset(self):
         if school_id:
             queryset = queryset.filter(school_id=school_id)
 
+        committee_id = query_params.get('committee_id', None)
+        if committee_id:
+            queryset = queryset.filter(assignment__committee_id=committee_id)
+
         return queryset
 
+    def patch(self, request, *args, **kwargs):
```
tdowds: This method is defined in `ListUpdateModelMixin`. No need to define it here since it's inherited.

huxley/api/permissions.py
```diff
@@ -67,56 +67,78 @@ def has_permission(self, request, view):
...
-                user.school.id == delegate.school.id)
+        if request.method in permissions.SAFE_METHODS:
+            return user_is_chair(request, view) or user_is_advisor(request, view)
 
+        return False
+        
 
-class AssignmentListPermission(permissions.BasePermission):
+class DelegateDetailPermission(permissions.BasePermission):
+    '''Accept requests to create, get, and update a delegate from the
```
tdowds: No create in `DelegateDetail`. It's Retrieve, Update and Destroy

tdowds:
This is some really solid work!

huxley/api/permissions.py
```diff
@@ -67,56 +67,78 @@ def has_permission(self, request, view):
...
-            if method == 'POST':
-                return int(request.data['school']) == school_id
-            else:
-                delegate_ids = [delegate['id'] for delegate in request.data]
-                return not Delegate.objects.filter(id__in=delegate_ids).exclude(school_id=school_id).exists()
+            return user_is_chair(request, view) or user_is_advisor(request, view)
+
+        if method == 'POST':
+            school_id = user.school_id
+            return (user.is_advisor() and school_id and
```
kmeht: I think we can write the `user_is_advisor` function to be a bit more general by accepting a `school_id` parameter, so that we don't have to do all these checks everywhere we need to check for an advisor.

huxley/api/permissions.py
```diff
@@ -67,56 +67,78 @@ def has_permission(self, request, view):
...
+            school_id = user.school_id
+            return (user.is_advisor() and school_id and 
+                    int(request.data['school']) == school_id)
+
+        if method in ('PUT', 'PATCH'):
+            delegate_ids = [delegate['id'] for delegate in request.data]
+            delegates = Delegate.objects.filter(id__in=delegate_ids)
+            if user.is_chair():
+                committee_id = user.committee_id
+                return (committee_id and
```
kmeht: Must we check if this is set? It should be an invariant in the app that a Chair user must have a committee, etc. Or is that not true, e.g. officers may not have one?

huxley/api/permissions.py
```diff
@@ -67,56 +67,78 @@ def has_permission(self, request, view):
...
+            delegate_ids = [delegate['id'] for delegate in request.data]
+            delegates = Delegate.objects.filter(id__in=delegate_ids)
+            if user.is_chair():
+                committee_id = user.committee_id
+                return (committee_id and 
+                        not delegates.exclude(assignment__committee_id=committee_id).exists())
+
+            if user.is_advisor():
+                school_id = user.school_id
+                return (school_id and
```
kmeht: At least for advisors, they must have a school ID or something much bigger has gone wrong.

huxley/api/views/delegate.py
```diff
@@ -1,19 +1,19 @@
-# Copyright (c) 2011-2016 Berkeley Model United Nations. All rights reserved.
+# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
```
kmeht: wat

huxley/api/permissions.py
```diff
@@ -63,65 +62,94 @@ def has_permission(self, request, view):
...
 
     def has_permission(self, request, view):
-        if request.user.is_superuser:
+        user = request.user
+        if user.is_superuser:
             return True
 
-        if request.method in permissions.SAFE_METHODS:
-            return user_is_advisor(request, view)
+        delegate_id = view.kwargs.get('pk', None)
```
kmeht: I don't think we need this `try/accept`, the delegate ID is a required parameter of the DelegateDetail URI. We can just get away with saying
```python
delegate_id = view.kwargs['pk']
```

tdowds:
This looks really good to me! No comments

huxley/api/permissions.py
m-j-mcdonald: My main line of thought wasn't to catch if a delegate id isn't provided but if a nonexistent one is. E.g., a chair & advisor login at the same time, the advisor deletes a delegate, then the chair tries to update the delegate. It seems like an outside chance so I thought it'd be good to check for it. @kmeht

m-j-mcdonald: The kwargs.get() syntax was just holdover from how the old permissions class did it

kmeht: Okay, I can buy that. In that case, I'd say we still remove the `try/except`. If the delegate doesn't exist, we should throw and return an error to the user; it's still not a permission failure.

PR #544, opened 2016-12-20
Uses flux for Assignment, Country, Committee and CurrentUser's School updates

tdowds:
This PR uses flux for data updates on Assignments, Countries, Committees and the CurrentUser's School in AdvisorAssignmentView (#523). I wanted to get some eyes on this diff before going into the other components (e.g. AdvisorProfileView and RegistrationView) and updating them to reflect these changes. Additionally, I need to update the tests for these stores as well.


huxley/www/js/actions/AssignmentActions.js
```diff
@@ -0,0 +1,27 @@
...
+
+var AssignmentActions = {
+  assignmentsFetched() {
+    Dispatcher.dispatch({
+      actionType: ActionConstants.ASSIGNMENTS_FETCHED
+    });
+  },
+
+  updateAssignment(assignmentID, delta) {
+  	Dispatcher.dispatch({
```
kmeht: consistently use spaces to indent your code

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -230,51 +236,17 @@ var AdvisorAssignmentsView = React.createClass({
     var confirm = window.confirm("By pressing okay you are committing to the financial responsibility of each assignment. Are you sure you want to finalize assignments?");
     var school = CurrentUserStore.getCurrentUser().school;
     if (confirm) {
-      this.setState({loading: true});
-      ServerAPI.updateSchool(school.id, {assignments_finalized: true})
-        .then(this._handleFinalizedSuccess, this._handleError);
+      var delta = {assignments_finalized: true};
```
kmeht: You can just inline this into the function call, no need for the separate binding:
```js
CurrentUserActions.updateSchool(school.id, {
  assignments_finalized: true,
});
```

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -230,51 +236,17 @@ var AdvisorAssignmentsView = React.createClass({
...
 
   _handleAssignmentDelete: function(assignment) {
     var confirm = window.confirm("Are you sure you want to delete this assignment?");
     if (confirm) {
-      this.setState({loading: true});
-      ServerAPI.updateAssignment(assignment.id, {rejected: true}).then(
-        this._handleAssignmentDeleteSuccess.bind(this, assignment.id),
-        this.handleError
-      );
+      var delta = {rejected: true};
```
kmeht: ditto

huxley/www/js/actions/CurrentUserActions.js
```diff
@@ -29,10 +29,11 @@ var CurrentUserActions = {
   },
 
 
-  updateSchool(school) {
+  updateSchool(schoolID, delta) {
     Dispatcher.dispatch({
       actionType: ActionConstants.UPDATE_SCHOOL,
-      school: school
+      schoolID: schoolID,
+      delta: delta
```
kmeht: tiny nit: trailing comma

huxley/www/js/stores/AssignmentStore.js
```diff
@@ -1,32 +1,57 @@
...
+  getAssignments(schoolID) {
+    if (_schoolsAssignments[schoolID]) {
+      return _schoolsAssignments[schoolID];
     }
-    if (callback) {
-      _assignmentPromises[schoolID].then(callback);
-    }
-    return _assignmentPromises[schoolID];
+
+    ServerAPI.getAssignments(schoolID).then(value => {
```
kmeht: If we call `AssignmentStore.getAssignments()` twice in a row, we will have sent this async request twice, which we don't want. We need a way to track that a request is in-flight.

Also, it doesn't appear that this does any error handling, e.g. if it fails there's no retry logic.

huxley/www/js/stores/CommitteeStore.js
```diff
@@ -5,38 +5,42 @@
...
-    }
-    return _committeePromise;
-  }
 
-  getSpecialCommittees(callback) {
-    var p = this.getCommittees().then((committees) => {
-      return committees.filter((committee) => committee.special);
+    ServerAPI.getCommittees().then(value => {
+      _committees = value;
+      CommitteeActions.committeesFetched();
```
kmeht: This action should contain the `value`, and the Store should only set `_committees` from the data passed into the action.

huxley/www/js/stores/CountryStore.js
```diff
@@ -1,32 +1,42 @@
...
+      return _countries;
     }
-    if (callback) {
-      _countryPromise.then(callback);
-    }
-    return _countryPromise;
+
+    ServerAPI.getCountries().then(value => {
+      _countries = value;
+      CountryActions.countriesFetched();
```
kmeht: Same as with Committees.

huxley/www/js/stores/AssignmentStore.js
```diff
@@ -1,32 +1,57 @@
...
+      _schoolsAssignments[schoolID] = value;
+      for (const assignment of value) {
+        _assignments[assignment.id] = assignment;
+      }
+      AssignmentActions.assignmentsFetched();
```
kmeht: Same with Committees and Country, the data should be passed into the action, and the value only updated in response to an action in `__onDispatch`.

Seems a little convoluted now, but it'll be clearer as we continue to iterate on the design.

huxley/www/js/stores/CountryStore.js
tdowds: Should I make a similar change in the DelegateStore?

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -230,51 +236,17 @@ var AdvisorAssignmentsView = React.createClass({
...
+      AssignmentActions.updateAssignment(assignment.id, delta);
     }
-  },
-
-  _handleSave: function(event) {
```
tdowds: Should not have deleted this

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -230,51 +236,24 @@ var AdvisorAssignmentsView = React.createClass({
...
-      ServerAPI.updateSchool(school.id, {assignments_finalized: true})
-        .then(this._handleFinalizedSuccess, this._handleError);
+      CurrentUserActions.updateSchool(school.id, {
+        assignments_finalized: true,
+      });
     }
   },
 
   _handleAssignmentDelete: function(assignment) {
     var confirm = window.confirm("Are you sure you want to delete this assignment?");
```
kmeht: This is for another diff, but the terminology could be better. "Decline" an assignment, perhaps?

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -195,6 +193,14 @@ var AdvisorAssignmentsView = React.createClass({
     );
   },
 
+  _makeMap: function(objectList) {
```
kmeht: I think it's much more likely that we want the countries and committees as a lookup table than a list. So, I think they should be stored as an object in the Store (i.e, this list-to-map conversion happens when the countries/committees are fetched).

This will also save us calling `_makeMap` all over the place.

huxley/www/js/components/RegistrationView.js
```diff
@@ -39,10 +39,12 @@ var RegistrationView = React.createClass({
   },
 
   getInitialState: function() {
+    var committees = CommitteeStore.getSpecialCommittees();
```
kmeht: Nit: these don't need to be variables, the calls can just be inlined into the returned object.

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -32,57 +33,54 @@ var AdvisorAssignmentsView = React.createClass({
 
   getInitialState: function() {
     var user = CurrentUserStore.getCurrentUser();
+    var assignments = AssignmentStore.getAssignments(user.school.id).filter(assignment => !assignment.rejected);
+    var committees = this._makeMap(CommitteeStore.getCommittees());
```
kmeht: This (and the three below) don't need to be variables, can inline

huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -230,51 +236,24 @@ var AdvisorAssignmentsView = React.createClass({
...
   },
 
   _handleSave: function(event) {
     var school = CurrentUserStore.getCurrentUser().school;
     DelegateActions.updateDelegates(school.id, this.state.delegates);
-  },
-
-  _handleFinalizedSuccess: function(response) {
-    CurrentUserActions.updateSchool(response);
-    this.setState({loading: false});
```
kmeht: This is definitely for another PR (and not necessarily hi-pri), but just putting it on your radar:

Perhaps there should be some kind of global "loading" state. Since updates are optimistic, it's possible that a user could make a change and then close the browser before the change is reflected in the server (possibly due to poor network conditions or something). In that case, we want to show a dialog prompting the user to wait a little longer while the request completes.

huxley/www/js/components/RegistrationView.js
```diff
@@ -90,12 +92,20 @@ var RegistrationView = React.createClass({
   },
 
   componentDidMount: function() {
-    CountryStore.getCountries(function(countries) {
-      this.setState({countries: countries});
-    }.bind(this));
-    CommitteeStore.getSpecialCommittees(function(committees) {
+    this._committeesToken = CommitteeStore.addListener(() => {
```
kmeht: I'm very much looking forward to getting all of our Stores onto standard Flux so that we can use FluxContainer to get rid of all this boilerplate.

huxley/www/js/components/RegistrationView.js
```diff
@@ -90,12 +92,20 @@ var RegistrationView = React.createClass({
...
+      var committees = CommitteeStore.getSpecialCommittees();
```
kmeht: not gonna comment on every instance of this, but these don't need to be variables

huxley/www/js/stores/CurrentUserStore.js
```diff
@@ -33,6 +34,12 @@ class CurrentUserStore extends Store {
     return super.addListener(callback);
   }
 
+  updateSchool(schoolID, delta) {
+    const school = {...this._currentUser.school, ...delta};
+    ServerAPI.updateSchool(schoolID, delta);
+    this._currentUser.school = school;
```
kmeht: Remember not to mutate:
```js
const user = this._currentUser;
this._currentUser = {
  ...user,
  school: {...user.school, ...delta},
};
```
Adding the temp `user` binding here because it's shorter, lol.

It's a little clunky, but JS doesn't have builtin support for immutable objects yet.

huxley/www/js/stores/__tests__/CurrentUserStore-test.js
```diff
@@ -73,4 +75,29 @@ describe('CurrentUserStore', function() {
...
+      actionType: ActionConstants.UPDATE_SCHOOL,
+      schoolID: 1,
+      delta: delta
+    });
+    expect(callback).toBeCalled();
+    expect(ServerAPI.updateSchool).toBeCalledWith(1, delta);
+
+    mockUser = CurrentUserStore.getCurrentUser();
+    expect(mockUser.school.name).toEqual(delta.name);
+    expect(mockUser.school.assignments_finalized);
```
kmeht: this looks incomplete?

huxley/www/js/stores/__tests__/DelegateStore-test.js
```diff
@@ -31,66 +42,131 @@ describe('DelegateStore', () => {
...
+    var new_delegate = {id: 3, name: 'Trevor', email: "", school: mockSchoolID};
+    registerCallback({
+      actionType: ActionConstants.ADD_DELEGATE,
+      delegate: new_delegate
+    });
+    expect(callback).toBeCalled();
+
+    var delegates = DelegateStore.getDelegates(mockSchoolID);
+    expect(delegates.length).toEqual(3);
+    expect(new_delegate in delegates);
```
kmeht: Does `expect()` without any methods called on it do anything?

huxley/www/js/stores/__tests__/AssignmentStore-test.js
```diff
@@ -28,15 +41,53 @@ describe('AssignmentStore', () => {
...
+    registerCallback({
+      actionType: ActionConstants.UPDATE_ASSIGNMENT,
+      assignmentID: 1,
+      delta: {rejected: true}
+    });
+    expect(callback).toBeCalled();
+
+    expect(ServerAPI.updateAssignment).toBeCalled();
+    var assignments = AssignmentStore.getAssignments(mockSchoolID);
+    expect(assignments[0]['rejected']);
```
kmeht: here's another missing expectation

PR #545, opened 2016-12-22
removes delegate assignment on change of assignment's school field

tdowds:
In the case where external changes the school field of an assignment, any delegate previously assigned to that assignment would remain assigned to it. Due to whatever unpredictable behavior that may come from that, this diff ensures that no delegate is assigned to an assignment when that assignment's school field is changed. It also sets that assignment's rejected field to `False`.

Additionally, I noticed a small bug in the `update_assignments` method of the Assignment model, and updated the tests to reflect that.

### DIFF
--- a/huxley/core/models.py
+++ b/huxley/core/models.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+# Copyright (c) 2011-2016 Berkeley Model United Nations. All rights reserved.
 # Use of this source code is governed by a BSD License (see LICENSE).
 
 import json
@@ -306,7 +306,7 @@ class Assignment(models.Model):
             deletions.append(assignment_data['id'])
 
         for committee, country, school, rejected in new_assignments:
-            key = (committee, country)
+            key = (committee.id, country.id)
             if key in assigned:
                 # Make sure that the same committee/country pair is not being
                 # given to more than one school in the upload
@@ -361,6 +361,20 @@ class Assignment(models.Model):
 
         return failed_assignments
 
+    @classmethod
+    def update_assignment(cls, **kwargs):
+        assignment = kwargs['instance']
+        # Check that this assignment already exists
+        if assignment.id:
+            old_assignment = cls.objects.get(id=assignment.id)
+            # Check if school changed
+            if not assignment.school.id == old_assignment.school.id:
+                assignment.rejected = False
+                delegates = Delegate.objects.filter(assignment_id=old_assignment.id)
+                for delegate in delegates:
+                    delegate.assignment = None
+                    delegate.save()
+
     def __unicode__(self):
         return self.committee.name + " : " + self.country.name + " : " + (
             self.school.name if self.school else "Unassigned")
@@ -369,6 +383,7 @@ class Assignment(models.Model):
         db_table = u'assignment'
         unique_together = ('committee', 'country')
 
+pre_save.connect(Assignment.update_assignment, sender=Assignment)
 
 class CountryPreference(models.Model):
     school = models.ForeignKey(School)
--- a/huxley/core/models.py
+++ b/huxley/core/models.py
@@ -371,7 +371,7 @@ class Assignment(models.Model):
             return
 
         old_assignment = cls.objects.get(id=assignment.id)
-        if not assignment.school.id == old_assignment.school.id:
+        if assignment.school_id != old_assignment.school_id:
             assignment.rejected = False
             Delegate.objects.filter(assignment_id=old_assignment.id).update(assignment=None)
 
--- a/huxley/core/models.py
+++ b/huxley/core/models.py
@@ -363,17 +363,17 @@ class Assignment(models.Model):
 
     @classmethod
     def update_assignment(cls, **kwargs):
+        '''Ensures that when an assignment's school field changes,
+           any delegates assigned to that assignment are no longer
+           assigned to it and that its rejected field is false.'''
         assignment = kwargs['instance']
-        # Check that this assignment already exists
-        if assignment.id:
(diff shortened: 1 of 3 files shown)


huxley/core/tests/models/test_models.py
```diff
@@ -174,10 +178,29 @@ def test_update_assignments(self):
...
+        s1 = TestSchools.new_school(name='S1')
+        s2 = TestSchools.new_school(name='S2')
+        a = TestAssignments.new_assignment(school=s1, rejected=True)
+        d1 = TestDelegates.new_delegate(school=s1, assignment=a)
+        d2 = TestDelegates.new_delegate(school=s1, assignment=a)
+
+        a.school = s2
+        a.save()
+
+        delegates = Delegate.objects.all()
```
kmeht: You can also refer back to the delegates from assignments, using the `related_name` param of the ForeignKey field on delegates:
- https://docs.djangoproject.com/en/1.8/topics/db/queries/#following-relationships-backward
- https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.related_name

So you can do something with the `a.delegates` QuerySet. I believe QuerySet has an `exists` or `count` operation, but you can double-check the docs.

huxley/core/models.py
```diff
@@ -361,6 +361,20 @@ def remove(assignment_data):
 
         return failed_assignments
 
+    @classmethod
+    def update_assignment(cls, **kwargs):
+        assignment = kwargs['instance']
+        # Check that this assignment already exists
+        if assignment.id:
```
kmeht: nit: I'd return early to prevent too much indentation with an `if not assignment.id` instead.

huxley/core/models.py
```diff
@@ -361,6 +361,20 @@ def remove(assignment_data):
 
         return failed_assignments
 
+    @classmethod
+    def update_assignment(cls, **kwargs):
+        assignment = kwargs['instance']
+        # Check that this assignment already exists
```
kmeht: Redundant comment, remove

huxley/core/models.py
```diff
@@ -361,6 +361,20 @@ def remove(assignment_data):
...
+        if assignment.id:
+            old_assignment = cls.objects.get(id=assignment.id)
+            # Check if school changed
```
kmeht: Redundant comment, remove. Instead, add a brief docstring to this function.

huxley/core/models.py
```diff
@@ -361,6 +361,20 @@ def remove(assignment_data):
...
+            if not assignment.school.id == old_assignment.school.id:
+                assignment.rejected = False
+                delegates = Delegate.objects.filter(assignment_id=old_assignment.id)
```
kmeht: You can do this more efficiently (and tersely) with the `update` method of a QuerySet, I believe. However, double-check that there aren't any `pre_save`/`post_save` signals attached to Delegate, because `QuerySet.update` doesn't trigger them.

https://docs.djangoproject.com/en/1.8/ref/models/querysets/#update

huxley/core/models.py
```diff
@@ -306,7 +306,7 @@ def remove(assignment_data):
             deletions.append(assignment_data['id'])
 
         for committee, country, school, rejected in new_assignments:
-            key = (committee, country)
+            key = (committee.id, country.id)
```
kmeht: brutal. was this causing all the assignments to be deleted? Why didn't our tests catch this?

tdowds: So I think the intended behavior is to have all the assignments be deleted. If you look towards the end of the method, we iterate over all assignments that weren't changed and then delete them. However, the reason why this change didn't necessarily have any effect is because when we try accessing `assigned_dict` with this key, it won't exist in it (because the keys are all IDs and not model instances). So we end up adding the assignment, and then deleting the old one later because it seems like we never updated it.

huxley/core/models.py
tdowds: The Delegate's overridden save method isn't called, but I don't think it matters because that's ensuring that the delegate's school field and the delegate's assignment's school field are the same. However if the delegate doesn't have an assignment anymore then there's no point in checking that.

huxley/core/models.py
```diff
@@ -361,6 +361,20 @@ def remove(assignment_data):
...
+    def update_assignment(cls, **kwargs):
+        '''Ensures that when an assignment's school field changes,
+           any delegates assigned to that assignment are no longer
+           assigned to it and that its rejected field is false.'''
+        assignment = kwargs['instance']
+        if not assignment.id:
+            return
+
+        old_assignment = cls.objects.get(id=assignment.id)
+        if not assignment.school.id == old_assignment.school.id:
```
kmeht: two things:
- `foo != bar` is more readable than `not foo == bar`
- Accessing the `school` property of the assignment triggers a DB fetch of the school object. Instead, just use the `school_id` property, since that's all we need: https://docs.djangoproject.com/en/1.8/topics/db/optimization/#use-foreign-key-values-directly

huxley/core/tests/models/test_models.py
```diff
@@ -174,10 +178,27 @@ def test_update_assignments(self):
...
+    def test_update_assignment(self):
+        '''Tests that when an assignment changes schools, its rejected
+           field is set to False and any delegates assigned to it are
+           no longer assigned to it.'''
+        s1 = TestSchools.new_school(name='S1')
+        s2 = TestSchools.new_school(name='S2')
+        a = TestAssignments.new_assignment(school=s1, rejected=True)
+        d1 = TestDelegates.new_delegate(school=s1, assignment=a)
+        d2 = TestDelegates.new_delegate(school=s1, assignment=a)
+        self.assertEquals(a.delegates.count(), 2)
```
kmeht: As a sanity check, might want to assert that `rejected` is True here.

Issue #547, opened 2016-12-24
Merge SchoolAssignments/SchoolDelegates views into AssignmentList/DelegateList, respectively.


kmeht:
As mentioned in [the comment on the other PR](https://github.com/bmun/huxley/pull/543/files#r91011495), these don't need to be separate views. Instead, they can just be filters on the main model view, where we implement the `get_queryset()` method and just fall back to `all()` if the filter isn't provided.

Care should be taken to ensure that the permissions remain correct. My first draft would be:
- superusers can get all delegates and assignments
- advisors can get all delegates/assignments for their school
- superusers can create delegates/assignments
- advisors can create delegates for their school, but NOT assignments

This may require rethinking how we implement our permissions, since they're a bit more complex (changing based on GET or POST, etc.).

kmeht:
I'll work on the first one as an example.

PR #548, opened 2016-12-27
Merge SchoolAssignments into AssignmentList (#547)

kmeht:
This should not be a separate API view; it should instead be a filter on AssignmentList. This merges the two and deletes SchoolAssignments.

Permissions will need to be a bit more complex, so instead of completely general-purpose Permission classes, this proposes a Permission class per-view (when needed), which then delegates to common functions for code reuse.

Tests are updated to verify the filtering behavior, and I inspected the request/response manually in the JS console in the browser.


kmeht:
Yep, should be fairly simple to follow this PR as an example for other similar use cases.

PR #549, opened 2016-12-29
merges SchoolDelegates into DelegateList

tdowds:
This is for the delegate side of #547. This PR removes SchoolDelegates and transfers the endpoint for getting delegates by `school_id` and updating delegates in bulk to DelegateList. Unit tests are updated accordingly, and I tested it out manually. Some important considerations from this PR include:

1. What kind of permissions do we want for bulk delegate updates?
2. Updating ListUpdateModelMixin to be model agnostic
3. Not using the auto formatter for our python code

For the first point, it currently allows anyone to create delegates and update them in bulk. It would probably be worth investigating how to tighten up those permissions.

For the second point, the ListUpdateModelMixin currently only updates delegates in bulk. However we should adjust it so that any model can be used by it. Additionally, I had to move the `put` and `patch` methods into the mixin because I kept getting a `405` error saying `PATCH method not allowed`. For consistency we might want to create our own custom bulk update view that inherits from ListUpdateModelMixin like how they do in DRF.

For the third point, I commented out the autoformatter because it couldn't find the file I deleted and threw an exception for it.


huxley/api/permissions.py
```diff
@@ -91,6 +91,23 @@ def has_permission(self, request, view):
         return False
 
 
+class DelegateListPermission(permissions.BasePermission):
+
+    def has_permission(self, request, view):
+        user = request.user
```
kmeht: But feel free to leave it if we end up using the user for the checks below :)

huxley/api/permissions.py
```diff
@@ -91,6 +91,23 @@ def has_permission(self, request, view):
...
+        if user.is_superuser:
+            return True
+
+        method = request.method
+        if method == 'POST' or method == 'PUT' or method == 'PATCH':
+            return True
```
kmeht: We should probably not let anybody update whatever delegates they want. Checking for it is a little less than straightforward though. Basically, we want to check that there is no delegate for which the `school_id` is not equal to the advisor's `school_id`. Something like this might work:
```python
school_id = request.user.is_authenticated() and request.user.school_id
if not school_id:
    return False
# Get delegate_ids via list comprehension...
return not Delegate.objects.filter(id__in=delegate_ids).exclude(school_id=school_id).exists()
```

huxley/api/tests/test_delegate.py
```diff
@@ -200,3 +202,283 @@ def test_superuser(self):
...
+        response = self.get_response(params={'school_id': self.school.id})
+        self.assert_delegate_equal(response)
+
+    def test_other_user(self):
+        '''It rejects a request from another user.'''
+        user2 = models.new_user(username='another', password='user')
+        models.new_school(user=user2)
+        self.client.login(username='another', password='user')
+
+        response = self.get_response(self.school.id)
```
kmeht: Need to pass `params` dictionary

huxley/api/tests/test_delegate.py
```diff
@@ -200,3 +202,283 @@ def test_superuser(self):
...
+        self.delegate2 = models.new_delegate(
+            assignment=self.assignment2,
+            name='Trevor Dowds',
+            email='t@dowds.com',
+            summary='Good!'
+        )
+
+    def test_anonymous_user(self):
+        '''It rejects a request from an anonymous user.'''
+        response = self.get_response(self.school.id)
```
kmeht: params

huxley/api/tests/test_delegate.py
```diff
@@ -200,3 +202,283 @@ def test_superuser(self):
...
+
+        response = self.get_response(self.school.id)
+        self.assertPermissionDenied(response)
+
+    def test_superuser(self):
+        '''It returns the delegates for a superuser.'''
+        models.new_superuser(username='test', password='user')
+        self.client.login(username='test', password='user')
+
+        response = self.get_response(self.school.id)
```
kmeht: params

huxley/api/tests/test_delegate.py
```diff
@@ -200,3 +202,283 @@ def test_superuser(self):
...
         )
+
+
+class DelegateListGetTestCase(tests.ListAPITestCase):
+    url_name = 'api:delegate_list'
+
+    def setUp(self):
+        self.user = models.new_user(username='regular', password='user')
+        self.school = models.new_school(user=self.user)
+        self.country = models.new_country()
```
kmeht: Don't think we need to explicitly construct the country or the committees, we don't care about them in this test

huxley/api/tests/test_delegate.py
```diff
@@ -200,3 +202,283 @@ def test_superuser(self):
...
+        )
+        self.assignment2 = models.new_assignment(
+            committee=self.committee2,
+            country=self.country,
+            school=self.school,
+        )
+        self.delegate1 = models.new_delegate(
+            assignment=self.assignment1,
+        )
+        self.delegate2 = models.new_delegate(
```
kmeht: It'd help to have a third delegate who's from a different school to make sure the filtering works.

huxley/api/permissions.py
```diff
@@ -91,6 +95,35 @@ def has_permission(self, request, view):
...
+
+        method = request.method
+        if method in permissions.SAFE_METHODS:
+            return user_is_advisor(request, view)
+
+        if method in ('POST', 'PUT', 'PATCH'):
+            school_id = user.is_authenticated() and user.school_id
+            if not school_id:
+                False
```
kmeht: `return False`

huxley/api/permissions.py
```diff
@@ -91,6 +95,35 @@ def has_permission(self, request, view):
...
+
+            if method == 'POST':
+                return int(request.data['school']) == school_id
+            else:
+                data = request.data
+                if isinstance(data, QueryDict):
```
kmeht: Why do we need this hack again? I remember you telling me, but I forgot the details.

huxley/api/permissions.py
```diff
@@ -91,6 +95,35 @@ def has_permission(self, request, view):
         return False
 
 
+class DelegateListPermission(permissions.BasePermission):
+
+    def has_permission(self, request, view):
```
kmeht: This is complex enough that it warrants a docstring explaining the permission.

huxley/api/permissions.py
```diff
@@ -1,10 +1,14 @@
...
 # Use of this source code is governed by a BSD License (see LICENSE).
 
-from rest_framework import permissions
+import json
+
+from django.http import QueryDict
 
 from huxley.core.models import Assignment, Delegate
 
+from rest_framework import permissions
```
kmeht: tiny nit: this goes directly under the `django.http` import, since they're both libraries

huxley/api/permissions.py
tdowds: Without it our unit tests wouldn't pass. When I manually test the data comes in as a QueryDict. However when I run the unit tests the data comes in as a list of objects.

kmeht: Thanks for the explanation! I'll open a task to fix this.

PR #552, opened 2016-12-30
uses flux to update user

tdowds:
Should be the last part of #523


huxley/www/js/stores/CurrentUserStore.js
```diff
@@ -36,13 +38,32 @@ class CurrentUserStore extends Store {
...
-    const user = this._currentUser.school;
+    const user = this._currentUser;
     this._currentUser = {
       ...user,
       school: {...user.school, ...delta},
     };
   }
 
+  updateUser(userID, delta) {
+    $.ajax({
```
kmeht: Add this as a function on ServerAPI, instead of doing the raw async request here.

huxley/www/js/stores/CurrentUserStore.js
```diff
@@ -36,13 +38,32 @@ class CurrentUserStore extends Store {
...
+      type: 'PATCH',
+      url: '/api/users/' + userID,
+      data: JSON.stringify(delta),
+      contentType: 'application/json',
+    });
+
+    const user = {
+      ...this._currentUser,
+      first_name: delta.first_name,
```
kmeht: This assumes that the delta will always have first and last name, which is true for now, but may not always be. We don't need to change anything right now, but just keep it in mind.

huxley/www/js/stores/__tests__/CurrentUserStore-test.js
```diff
@@ -100,4 +100,33 @@ describe('CurrentUserStore', function() {
...
+    var delta = {first_name: 'Trev', school: {primary_email: "trev@d.com"}};
+    registerCallback({
+      actionType: ActionConstants.UPDATE_USER,
+      userID: 2,
+      delta: delta
+    });
+    expect(callback).toBeCalled();
+
+    mockUser = CurrentUserStore.getCurrentUser();
+    expect(mockUser.first_name).toEqual(delta.first_name);
```
kmeht: If you assert on the `last_name` too, I suspect that it will be undefined (based on the comment I left above).

huxley/www/js/stores/CurrentUserStore.js
```diff
@@ -1,10 +1,12 @@
 /**
- * Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+ * Copyright (c) 2011-2016 Berkeley Model United Nations. All rights reserved.
  * Use of this source code is governed by a BSD License (see LICENSE).
  */
 
 'use strict';
 
+var $ = require('jquery');
```
kmeht: remove, once the request is in the serverAPI

PR #553, opened 2016-12-30
Reincorporates error handling using flux

tdowds:
The approach to do this is to pass the original error handlers into Actions where those handlers will eventually be called by stores.

For this particular PR I used AdvisorRosterView as an example for what I will do for the remaining views. In it I was unsure of how necessary it was to handle errors when deleting a delegate. I added it for completeness. 

Additionally, something that might be worth considering is when an error is returned from the server, should we automatically roll back those changes in the store? This would probably require firing another action so that the roll back is emitted to the view. If we don't then to the user the change appears implemented (minus the error message thrown) until the page is refreshed.

### DIFF
--- a/huxley/www/js/actions/DelegateActions.js
+++ b/huxley/www/js/actions/DelegateActions.js
@@ -9,10 +9,11 @@ var ActionConstants = require('constants/ActionConstants');
 var Dispatcher = require('dispatcher/Dispatcher');
 
 var DelegateActions = {
-  deleteDelegate(delegateID) {
+  deleteDelegate(delegateID, error) {
     Dispatcher.dispatch({
       actionType: ActionConstants.DELETE_DELEGATE,
-      delegateID: delegateID
+      delegateID: delegateID,
+      error: error,
     });
   },
 
@@ -23,11 +24,12 @@ var DelegateActions = {
     });
   },
 
-  updateDelegate(delegateID, delta) {
+  updateDelegate(delegateID, delta, error) {
     Dispatcher.dispatch({
       actionType: ActionConstants.UPDATE_DELEGATE,
       delegateID: delegateID,
-      delta: delta
+      delta: delta,
+      error: error,
     });
   },
 
--- a/huxley/www/js/actions/DelegateActions.js
+++ b/huxley/www/js/actions/DelegateActions.js
@@ -9,11 +9,11 @@ var ActionConstants = require('constants/ActionConstants');
 var Dispatcher = require('dispatcher/Dispatcher');
 
 var DelegateActions = {
-  deleteDelegate(delegateID, error) {
+  deleteDelegate(delegateID, onError) {
     Dispatcher.dispatch({
       actionType: ActionConstants.DELETE_DELEGATE,
       delegateID: delegateID,
-      error: error,
+      onError: onError,
     });
   },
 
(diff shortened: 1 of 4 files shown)


huxley/www/js/actions/DelegateActions.js
```diff
@@ -9,10 +9,11 @@ var ActionConstants = require('constants/ActionConstants');
 var Dispatcher = require('dispatcher/Dispatcher');
 
 var DelegateActions = {
-  deleteDelegate(delegateID) {
+  deleteDelegate(delegateID, error) {
     Dispatcher.dispatch({
       actionType: ActionConstants.DELETE_DELEGATE,
-      delegateID: delegateID
+      delegateID: delegateID,
+      error: error,
```
kmeht: I'd name these `onError` to make it clear that they're callbacks.

huxley/www/js/actions/DelegateActions.js
```diff
@@ -23,11 +24,12 @@ var DelegateActions = {
     });
   },
 
-  updateDelegate(delegateID, delta) {
+  updateDelegate(delegateID, delta, error) {
```
kmeht: ditto

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -237,10 +237,17 @@ var AdvisorRosterView = React.createClass({
     });
   },
 
+  _handleDeleteError: function(response) {
+    window.confirm(
```
kmeht: I think an `alert` would be enough here.

huxley/www/js/stores/DelegateStore.js
```diff
@@ -28,8 +28,8 @@ class DelegateStore extends Store {
     return [];
   }
 
-  deleteDelegate(delegateID) {
-    ServerAPI.deleteDelegate(delegateID);
+  deleteDelegate(delegateID, error) {
+    ServerAPI.deleteDelegate(delegateID).then(null, error);
```
kmeht: You can use the `catch` method of Promise, which only handles errors.

huxley/www/js/stores/DelegateStore.js
```diff
@@ -40,9 +40,9 @@ class DelegateStore extends Store {
     _schoolsDelegates[delegate.school] = [..._schoolsDelegates[delegate.school], delegate];
   }
 
-  updateDelegate(delegateID, delta) {
+  updateDelegate(delegateID, delta, error) {
     const delegate = {..._delegates[delegateID], ...delta};
-    ServerAPI.updateDelegate(delegateID, delegate);
+    ServerAPI.updateDelegate(delegateID, delegate).then(null, error);
```
kmeht: ditto

huxley/www/js/stores/DelegateStore.js
```diff
@@ -28,8 +28,8 @@ class DelegateStore extends Store {
     return [];
   }
 
-  deleteDelegate(delegateID) {
-    ServerAPI.deleteDelegate(delegateID);
+  deleteDelegate(delegateID, error) {
```
kmeht: `onError`

huxley/www/js/stores/DelegateStore.js
```diff
@@ -40,9 +40,9 @@ class DelegateStore extends Store {
     _schoolsDelegates[delegate.school] = [..._schoolsDelegates[delegate.school], delegate];
   }
 
-  updateDelegate(delegateID, delta) {
+  updateDelegate(delegateID, delta, error) {
```
kmeht: `onError`

PR #554, opened 2016-12-31
Set correct contentType on AJAX requests

### LINKED ISSUE #551: Fix weird QueryDict hack for bulk update API views
When sending lists as POST data, the data comes in as a QueryDict. However, when running tests, the data comes in as a list of objects. This requires us to hack around the problem by doing `QueryDict` instance checks. Instead, we should fix our test framework to supply the data in the same way that Django would normally.

kmeht:
This fixes the need to do the QueryDict hack because it correctly sets the content type on the request, which causes it to be parsed as JSON instead of url-encoded data.

This also adds the JSON stringification in ServerAPI, so we don't have to do it manually in the calls.

Resolves #551.

### DIFF
--- a/huxley/api/mixins.py
+++ b/huxley/api/mixins.py
@@ -4,12 +4,11 @@
 import json
 
 from django.db import transaction
-from django.http import QueryDict
+from rest_framework import status
+from rest_framework.response import Response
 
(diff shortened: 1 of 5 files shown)


kmeht:
This is actually broken, fix incoming.

PR #555, opened 2016-12-31
Upgrade to Django 1.9.12

patilpranay:
I've run all the tests and they appear to pass without any needed to update any of the other requirements, most likely because 1.9 itself is pretty old.

However, once we decide to upgrade to Django 1.10, we will have to take care of deprecation warnings due to code part of the Django rest framework.

In terms of new features in 1.9, there is now password validation and an 'on_commit' hook that performs actions after a database transaction. The 'on_commit' hook seems like something useful that we could make some of the code simpler. Thoughts?


kmeht:
Hah, that was easy :)

kmeht:
@patilpranay, I noticed some deprecation warnings in the console when running the server, e.g.

`/Users/kunalm/Sites/Django/huxley/huxley/core/admin/delegate.py:68: RemovedInDjango110Warning: django.conf.urls.patterns() is deprecated and will be removed in Django 1.10. Update your urlpatterns to be a list of django.conf.urls.url() instances instead.`

In future upgrade PRs, these should be fixed along with updating the version in `requirements.txt` :)

Issue #556, opened 2017-01-08
Browser compatibility


m-j-mcdonald:
Right now the ES8 function Object.values is used in several places, such as RegistrationView and CommitteeStore, but only Chrome & Firefox currently support it. This means for example that someone running IE would be unable to register. We should find a way to replace these using functionality from ES7.

kmeht:
Whoops! I didn't realize it was nonstandard still. Babel or core-js should be able to polyfill it.

Issue #505, opened 2016-08-28
ChairSummaryView component


tdowds:
Chairs should be able to leave summaries of their delegates at the end of conference. Need to have a textarea per delegation in which to leave comments. Delegates should be updated in bulk. 

Because this information will be seen by advisors we should protect ourselves by having save and finalization functionality. Therefore whenever a chair saves, their work is saved but it is not visible to advisors. Whenever a chair finalizes, their work will be published to advisors.

PR #561, opened 2017-01-29
#505 Chair Summary View

m-j-mcdonald:
This is the view where chairs can enter feedback for delegations (it's done on a country-level granularity). It introduces a published_summary on the delegate model so chairs can have a published and unpublished version of their comments. This addresses issue #505. Also, this addresses issue #556 quickly (it's just a two line addition so I tacked it on here).


huxley/www/js/components/RegistrationView.js
```diff
@@ -5,6 +5,9 @@
 
 'use strict';
 
+// Only necessary until ES2017 becomes standard
```
m-j-mcdonald: This addresses issue #556. It solves the issue for Safari I know, but I don't have access to IE to test it there (there's no reason it shouldn't work there, I'm just saying I couldn't test it).

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,185 @@
...
+    var country_assignments = this.state.country_assignments;
+    var delegates = [];
+    for (var country in country_assignments) {
+      delegates = delegates.concat(country_assignments[country]);
+    }
+    DelegateActions.updateCommitteeDelegates(committee, delegates);
+  },
+
+  _handlePublishSummaries(event) {
+    var confirm = window.confirm("Advisors will be able to see everything you have entered. Do you wish to continue?");
```
tdowds: I'd say let's be even more explicit here. Something like "By pressing ok, you are allowing advisors to read the summaries that you have written about their delegations. Please ensure there are no inappropriate comments or language in any of these summaries. If there is none and you are ready to publish your summaries to advisors, press 'ok' to continue."

I hate to be a stickler about language, but I want to make sure there is absolutely no way for any of our chairs to accidentally publish anything that our advisors would find inappropriate.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,185 @@
...
+      return (
+        <tr key={country}>
+          <td>
+            {Object.keys(countries).length ? countries[country].name : country}
+          </td>
+          <td>
+            <textarea
+              className="text-input"
+              style={{"width": "95%"}}
+              rows="3"
```
tdowds: Depending on feedback we get, this may or may not need to be one or two rows larger

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,185 @@
...
+    this._delegatesToken && this._delegatesToken.remove();
+    this._assignmentsToken && this._assignmentsToken.remove();
+  },
+
+  render() {
+    return (
+      <InnerView>
+        <h2>Summaries</h2>
+        <p>
+          Here you can provide feedback for the delegates. Note that publishing
```
tdowds: We should probably be more explicit here as well: "Please note that clicking the "Publish" button will allow advisors to read the summaries you have written about their delegates."

huxley/www/js/components/RegistrationView.js
kmeht: This should go in the entry file (`www/js/entry.js`). Notice that it already has a corejs polyfill in there (we don't want to require the entire polyfill to keep our JS size down). Determine which polyfill we need, perhaps it's just `core-js/es7`, and put that there.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,185 @@
...
+      assignments: AssignmentStore.getCommitteeAssignments(user.committee),
+      countries: CountryStore.getCountries(),
+      country_assignments: {},
+      delegates: DelegateStore.getCommitteeDelegates(user.committee),
+    };
+  },
+
+  componentWillMount() {
+    var user = CurrentUserStore.getCurrentUser();
+    if (!User.isChair(user)) {
```
kmeht: Do we have to check this here? It seems like there should be a more central place where we check if a user has permission to view a particular page. Did we never build that in? Can chairs go to `advisor/whatever` and see that view (even though it'll be broken)?

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -42,6 +42,7 @@ var ChairAttendanceView = React.createClass({
 
   componentDidMount() {
     var user = CurrentUserStore.getCurrentUser();
+    this._mapAssignments();
     this._delegatesToken = DelegateStore.addListener(() => {
```
m-j-mcdonald: This is a bug I noticed and decided to wrap it into this PR since I was making changes anyways

m-j-mcdonald:
screen shot 2017-02-05 at 1 09 44 pm

m-j-mcdonald:
This update accounts for the comments. Per suggestions from Trevor I created some new styling for the save buttons

huxley/www/js/components/ChairSummaryView.js
m-j-mcdonald: Right now I'm pretty sure the only thing stopping a chair from viewing an advisor specific page is that the code will break and be unable to load

huxley/www/js/components/ChairSummaryView.js
m-j-mcdonald: This is a resizable component, so I feel 3 rows is a good default height

tdowds:
To clarify my suggestions, the goal of introducing new styling was so that chairs wouldn't have to scroll all the way to the bottom of the page in order to save their summaries (especially for chairs that have a lot of delegations).

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -89,12 +96,16 @@ var ChairAttendanceView = React.createClass({
               </tbody>
             </table>
           </div>
-          <Button
-            color="green"
-            onClick={this._handleSaveAttendance}>
-            Confirm Attendance
-          </Button>
         </form>
+        <div className="foot-bar">
```
kmeht: UI-wise, I'm opposed to having these buttons in a footer; a footer isn't specific to the particular page, so it's confusing to have page-specific buttons there.

If the issue is that user's shouldn't have to scroll to click save/publish, we could just make the table scrollable itself on the page: http://stackoverflow.com/questions/9707397/making-a-div-vertically-scrollable-using-css

That way, the buttons can be always visible, but the table is still scrollable.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,205 @@
...
+    var newSummary = event.target.value;
+    var country_assignments = this.state.country_assignments;
+    var country_delegates = country_assignments[country];
+    for (var delegate of country_delegates) {
+      delegate.summary = newSummary;
+    }
+    this.setState({country_assignments: country_assignments});
+  },
+
+  _mapAssignments() {
```
kmeht: That we need this function in the first place is pretty gross. IMO, before any more new features are built, y'all need to iron out how state is stored on the client so all of this isn't necessary.\

One thing I'm noticing is that summaries are really attached to assignments, so why don't we store them as a map from assignment ID to summary in a `summaries` state property? Then, when the user hits save or publish, we then match up the assignments with their delegates, update those fields, and send it to the server.

This way, the function can be as follows, and only needs to be called when delegates change:
```js
const summaries = {};
  // since we're duplicating the state between the two delegates, there will be an overwrite here per delegate, but whatever.
  summaries[delegate.assignment] = delegate.summary;
}
this.setState({summaries});
```

m-j-mcdonald:
screen shot 2017-02-12 at 7 03 09 pm

This uses a scrollable div limited to take up 50% of the vertical space.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -41,20 +41,32 @@ var ChairSummaryView = React.createClass({
...
+    var delegates =  this.state.delegates;
+    var summaries = this.state.summaries;
+    for (var delegate of delegates) {
+      summaries[delegate.assignment] = delegate.summary;
+    }
+
     this._delegatesToken = DelegateStore.addListener(() => {
-      this.setState({delegates: DelegateStore.getCommitteeDelegates(user.committee)});
-      this._mapAssignments();
+      var delegates =  DelegateStore.getCommitteeDelegates(user.committee);
```
tdowds: extra space after equals?

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -90,90 +103,69 @@ var ChairSummaryView = React.createClass({
...
-              Save
-            </Button>
-            <Button
-              color="blue"
-              onClick={this._handlePublishSummaries}>
-              Publish
-            </Button>
-          </ul>
-        </div>
+
```
tdowds: remove extra space

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -0,0 +1,192 @@
...
+  renderSummaryRows() {
+    var assignments = this.state.assignments;
+    var summaries = this.state.summaries;
+    var assignmentIDs = Object.keys(summaries);
+    var countries = this.state.countries;
+    assignments = assignments.filter(a => assignmentIDs.indexOf(""+a.id) > -1);
+    return assignments.map(assignment => {
+      return (
+        <tr key={countries[assignment.country].id}>
```
tdowds: is `assignment.country` != `countries[assignment.country].id`?

tdowds:
A couple of small nits that I noticed, but overall looks solid to me. @kmeht any other suggestions?

Issue #566, opened 2017-02-09
Prevent Add Delegate modal from disappearing


tdowds:
I'm hearing complaints from advisors that the modal to add a delegate often disappears before they are able to add. I have seen this happen locally. It seems to happen when you login, navigate to the roster view, and then hit add delegate - the page then refreshes on its own.

The source of this issue is that in the stores we emit a FETCHED action after we successfully get the items of that store from the server. However if nothing comes back this FETCHED actions get emitted, the component refreshes, and then calls the store's getter, and seeing that there are no delegates, repeats this process. This is causing the modal in the AdvisorRosterView to disappear when the advisor has no delegates.

What is needed for this task is some way of differentiating between "having no delegates" and "haven't fetched yet", and updated unit tests to catch this case

PR #570, opened 2017-02-19
Fixed delegate store with no delegates

m-j-mcdonald:
This address issue #566 by adding a simple flag to the store to tell whether or not delegates have been fetched. The old version was abusing the existing data by inferring whether data had been fetched using the value in the store and matching no data to having not fetched data.

### DIFF
--- a/huxley/www/js/components/AdvisorAssignmentsView.js
+++ b/huxley/www/js/components/AdvisorAssignmentsView.js
@@ -81,7 +81,9 @@ var AdvisorAssignmentsView = React.createClass({
 
   render: function() {
     var finalized = CurrentUserStore.getFinalized();
+    var committees = this.state.committees;
     var conference = this.context.conference;
+    var countries = this.state.countries;
     return (
       <InnerView>
         <h2>Assignments</h2>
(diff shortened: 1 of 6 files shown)


huxley/www/js/components/AdvisorAssignmentsView.js
```diff
@@ -112,7 +114,11 @@ var AdvisorAssignmentsView = React.createClass({
                 </tr>
               </thead>
               <tbody>
-                {this.renderAssignmentRows()}
+                {
+                  Object.keys(committees).length > 0 && Object.keys(countries).length > 0 ?
```
m-j-mcdonald: This is a side issue I've been noticing that seemed to small for a separate PR. The old code was triggering some promise rejections that didn't seem to be affecting anything but this provides a guarantee.

huxley/www/js/stores/DelegateStore.js
```diff
@@ -100,6 +100,7 @@ class DelegateStore extends Store {
         var userID = CurrentUserStore.getCurrentUser().id;
         if(userID != _previousUserID) {
```
tdowds: space after the if keyword

PR #573, opened 2017-02-25
Added ordering of countries by id into chair view

m-j-mcdonald:
This just adds some sorting to the lists seen by chairs, making it easier to find a specific delegation


tdowds:
Would it be possible to sort by name instead of id? I don't think it's guaranteed that sorting by id corresponds to sorting alphabetically

m-j-mcdonald:
This should sort by country name now.

PR #574, opened 2017-02-26
Added feature to differntiate present from present and voting

m-j-mcdonald:
This adds a checkbox into DelegationAttendanceRow for tracking if a delegation is "voting". This is only visible to the chairs.


tdowds:
This looks great. In the next diff, let's try to move the table headers outside of the scrollable part of our table, so chairs don't get confused on which column they are on in case we add more columns.

PR #575, opened 2017-03-01
Change table headers not to scroll

m-j-mcdonald:
This moves the table headers for the Chair View out of the scrollable div. It's a little hackish; I couldn't find an elegant way to keep the tables together.


kmeht:
You could make a "ScrollableTable" React component with sticky headers taking inspiration from this SO answer: http://stackoverflow.com/a/25902860. This way, you wouldn't have to duplicate the code in both views, and it wouldn't be as hacky.

You'd be using React's event system instead of native JS, but the concept is more-or-less the same.

PR #579, opened 2017-03-07
Only save summaries that have been modified

m-j-mcdonald:
This makes it so that when a chair saves summaries, only the delegates that have had their summaries updated get sent to the backend. This helps with letting multiple chairs access the page at once, as now only changes made to the same delegates' summaries would get clobbered (versus before, where every summary for the committee would get clobbered).

### DIFF
--- a/huxley/www/js/components/ChairSummaryView.js
+++ b/huxley/www/js/components/ChairSummaryView.js
@@ -187,10 +187,14 @@ var ChairSummaryView = React.createClass({
     var committee = CurrentUserStore.getCurrentUser().committee;
     var delegates = this.state.delegates;
     var summaries = this.state.summaries;
+    var toSave = [];
     for (var delegate of delegates) {
-      delegate.summary = summaries[delegate.assignment];
+      if (delegate.summary != summaries[delegate.assignment]) {
+        delegate.summary = summaries[delegate.assignment];
+        toSave.push(delegate);
+      }
     }
-    DelegateActions.updateCommitteeDelegates(committee, delegates);
+    DelegateActions.updateCommitteeDelegates(committee, toSave);
     event.preventDefault();
   },
 
@@ -207,11 +211,16 @@ var ChairSummaryView = React.createClass({
       var committee = CurrentUserStore.getCurrentUser().committee;
       var delegates = this.state.delegates;
       var summaries = this.state.summaries;
+      var toPublish = [];
       for (var delegate of delegates) {
-        delegate.summary = summaries[delegate.assignment];
+        var summary = summaries[delegate.assignment];
+        if (delegate.summary != summary || delegate.published_summary != summary) {
+          delegate.summary = summaries[delegate.assignment];
+          delegate.published_summary = delegate.summary;
+          toPublish.push(delegate);
+        }
       }
-      delegates.forEach(delegate => delegate.published_summary = delegate.summary);
-      DelegateActions.updateCommitteeDelegates(committee, delegates);
+      DelegateActions.updateCommitteeDelegates(committee, toPublish);
       event.preventDefault();
     }
   },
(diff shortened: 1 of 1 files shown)


huxley/www/js/components/ChairSummaryView.js
```diff
@@ -187,10 +187,14 @@ var ChairSummaryView = React.createClass({
     var committee = CurrentUserStore.getCurrentUser().committee;
     var delegates = this.state.delegates;
     var summaries = this.state.summaries;
+    var toSave = [];
     for (var delegate of delegates) {
-      delegate.summary = summaries[delegate.assignment];
+      if (delegate.summary != summaries[delegate.assignment]) {
+        delegate.summary = summaries[delegate.assignment];
```
kmeht: Not a huge fan of directly mutating the state object. Do we use `delegate.summary` in the UI, or is that what `summaries` is for? If we don't, we can let this slide, but we should be working on making that separate model and moving summaries/attendance to it.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -207,11 +211,16 @@ var ChairSummaryView = React.createClass({
       var committee = CurrentUserStore.getCurrentUser().committee;
       var delegates = this.state.delegates;
       var summaries = this.state.summaries;
+      var toPublish = [];
       for (var delegate of delegates) {
-        delegate.summary = summaries[delegate.assignment];
+        var summary = summaries[delegate.assignment];
+        if (delegate.summary != summary || delegate.published_summary != summary) {
+          delegate.summary = summaries[delegate.assignment];
```
kmeht: Use the `summary` variable.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -207,11 +211,16 @@ var ChairSummaryView = React.createClass({
...
+          delegate.published_summary = summaries[delegate.assignment];
```
kmeht: ditto

m-j-mcdonald:
This now sends an updated copy of the delegate instead of the original, avoiding potential state issues. This also means that local copies of delegates will match what's in the backend at all times. To be honest I need to make this change elsewhere as well, there are a few places in ChairSummaryView and ChairAttendanceView the state is mutated improperly.

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -187,10 +187,15 @@ var ChairSummaryView = React.createClass({
     var committee = CurrentUserStore.getCurrentUser().committee;
     var delegates = this.state.delegates;
     var summaries = this.state.summaries;
+    var toSave = [];
     for (var delegate of delegates) {
-      delegate.summary = summaries[delegate.assignment];
+      var summary = summaries[delegate.assignment];
+      if (delegate.summary != summary) {
+        var update = {summary:summary};
+        toSave.push(Object.assign({}, delegate, update));
```
kmeht: We have the object spread transform on for this:
```js
toSave.push({...delegate, ...update});
```

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -207,11 +212,15 @@ var ChairSummaryView = React.createClass({
       var committee = CurrentUserStore.getCurrentUser().committee;
       var delegates = this.state.delegates;
       var summaries = this.state.summaries;
+      var toPublish = [];
       for (var delegate of delegates) {
-        delegate.summary = summaries[delegate.assignment];
+        var summary = summaries[delegate.assignment];
+        if (delegate.summary != summary || delegate.published_summary != summary) {
+          var update = {summary:summary, published_summary:summary};
+          toPublish.push(Object.assign({}, delegate, update));
```
kmeht: Ditto:
```js
toPublish.push({...delegate, ...update});
```

PR #587, opened 2017-04-24
Initial setup for LESS

kmeht:
This sets up the infrastructure for using LESS to compile our CSS, with the eventual goal of getting off of Django Pipeline. Changes in this PR:
- Upgrades to Webpack 2
- Sets up all the loaders in Webpack
- Converts one file as an example
- Changes the CSS style to more closely match JS (two-space indentation, no newline for first brace)

Notice that our CSS files are now colocated with their corresponding React components, instead of in their own directory. This seems fundamentally easier to understand.

Eventually, once we're all converted to LESS, we'll be able to take advantage of its features if we choose, like mixins and variables.


tdowds:
This looks solid to me from what I can tell (not too familiar with LESS and how everything is compiled with it). But I think it might help organize things more to throw our LESS files into its own directory instead of with our directories that contain our js files (like a `huxley/www/less`). Unless that defeats the purpose of what we're trying to do here, but that's just my opinion on what's easier to understand.

PR #589, opened 2017-05-06
Split JS into app/vendor bundles.

### LINKED ISSUE #508: Split JS into multiple bundles
In our app, there will be code that changes relatively infrequently (library code) and frequently (application code). These should be separate bundles, so that the library code can be cached by the browser and we don't have to keep serving the same bytes repeatedly. Webpack has an option to do this.

kmeht:
This separates our app code from our dependency code. Because dependencies change a lot less frequently than the app, this allows the users' browsers to cache the vendor bundle, so that it loads faster on subsequent requests.

One minor downside is that when we add/remove dependencies, we also need to update the list in the Webpack config.

Resolves #508.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,7 +11,11 @@ var plugins = [
   new webpack.EnvironmentPlugin([
     'NODE_ENV',
   ]),
-  new ExtractTextPlugin('bundle.css'),
+  new ExtractTextPlugin('huxley.css'),
+  new webpack.optimize.CommonsChunkPlugin({
+    name: 'vendor',
+    filename: 'vendor.js',
+  }),
 ];
 if (process.env.NODE_ENV === 'production') {
   plugins.push(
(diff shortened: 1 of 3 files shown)


kmeht:
Documentation for this: https://webpack.js.org/plugins/commons-chunk-plugin/#explicit-vendor-chunk

webpack.config.js
```diff
@@ -25,10 +29,23 @@ if (process.env.NODE_ENV === 'production') {
...
+      'react',
+      'react-dom',
+      'react-modal',
+      'react-router',
+    ],
+  },
   output: {
     path: STATIC_ROOT,
-    filename: 'bundle.js',
+    filename: '[name].js',
```
tdowds: Just a bit confused about how webpack works, but is this referring to vendor or huxley?

kmeht: @tdowds That's a placeholder for the name of whatever bundle gets outputted.

PR #590, opened 2017-05-07
Use CSS animation to shake instead of jQuery.

### LINKED ISSUE #337: Use animate.css for shake animations
Using jQuery UI just for simple shake animations is overkill. Let's use animate.css instead, which provides a bunch of configurable animations done in CSS: https://github.com/daneden/animate.css. Or, we could roll our own animations.

This will help break our dependency on jQuery, tracked in #207.

kmeht:
This ditches jQuery for the "shake" error animation throughout the app. Instead, we use CSS animations inspired by https://github.com/daneden/animate.css. There's a top level Shaker component, which provides a `shake` method through context. That way, any child can trigger a shake. Resolves #337.

### DIFF
--- a/huxley/www/js/components/AdvisorView.js
+++ b/huxley/www/js/components/AdvisorView.js
@@ -10,6 +10,7 @@ var ReactRouter = require('react-router');
 
 var NavTab = require('components/NavTab');
 var PermissionDeniedView = require('components/PermissionDeniedView');
+var Shaker = require('components/Shaker');
 var TopBar = require('components/TopBar');
 var User = require('utils/User');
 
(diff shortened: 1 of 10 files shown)


huxley/www/js/components/AdvisorView.js
```diff
@@ -32,21 +33,23 @@ var AdvisorView = React.createClass ({
...
-          </NavTab>
-          <NavTab href="/advisor/roster">
-            Delegates
-          </NavTab>
-          <NavTab href="/advisor/feedback">
-            Feedback
-          </NavTab>
-        </div>
-        {this.props.children}
+        <Shaker>
```
tdowds: If we want to use other features by animate.css, will that require us to define top level components like this as well?

huxley/www/js/components/Huxley.js
```diff
@@ -49,7 +50,11 @@ var Huxley = React.createClass({
   render: function() {
     var user = CurrentUserStore.getCurrentUser();
     if (User.isAnonymous(user)) {
-      return React.cloneElement(this.props.children, { user: user });
+      return (
+        <Shaker>
```
tdowds: Will this shake the whole screen or just the view that contains our forms (e.g. LoginView, RegistrationView)? And does it make a difference either way?

huxley/www/js/components/Shaker.js
```diff
@@ -0,0 +1,47 @@
...
+
+var React = require('react');
+var ReactDOM = require('react-dom');
+
+var cx = require('classnames');
+
+require('css/Shaker.less');
+
+var Shaker = React.createClass({
+  childContextTypes: {
```
m-j-mcdonald: Does this need to be wrapped into a component? A function stored in the utils folder might be simpler if it were possible, but I don't know how that would work with the CSS.

huxley/www/js/components/AdvisorView.js
kmeht: Probably. I haven't thought too much about it; I figure we'll cross that bridge when we get to it. One option might be renaming this component to be `Animatable` or something and have it support whichever animations we want to execute on it.

huxley/www/js/components/Shaker.js
kmeht: I thought about that, but it forces any component to know about its parent components. With this, we break the component abstraction a little with context, but the function isn't required to be passed-in, so the child components still don't force the parents to do anything.

huxley/www/js/components/Huxley.js
kmeht: This will only shake the actual div containing the content. Currently, it doesn't matter in practice because the outer views don't have any other UI like a top bar.

Issue #494, opened 2016-07-31
Log message for 400s


tdowds:
Mostly putting this here so I don't forget. But when an exception occurs we currently log the entire stack trace as the message of our log. However when 400s occur we don't log anything. From what I've seen 400s usually come with some kind message. We should log this message.

tdowds:
@chevans7 what you're going to want to do is in `huxley/core/middlewares.py` in the `process_response` method check the status of the response and if it's some kind of 400, log its content.

PR #592, opened 2017-05-19
logs 400 messages.

tdowds:
Related to #494, this fills the message field of our log object with whatever message is associated with our 400s responses.

### DIFF
--- a/huxley/core/middlewares.py
+++ b/huxley/core/middlewares.py
@@ -24,8 +24,7 @@ class LoggingMiddleware(object):
         if 'api' in request.path:
             logger = logging.getLogger('huxley.api')
             status_code = response.status_code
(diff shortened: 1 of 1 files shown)


huxley/core/middlewares.py
```diff
@@ -23,10 +23,13 @@ class LoggingMiddleware(object):
     def process_response(self, request, response):
         if 'api' in request.path:
             logger = logging.getLogger('huxley.api')
+            status_code = response.status_code
+            message = response.getvalue() \
+              if status_code >= 400 and status_code < 500 else ""
```
kmeht: Pretty sure Python supports composing comparison expressions, e.g.
```python
if 500 > status_code >= 500 else ''
```

Issue #559, opened 2017-01-21
Use Markdown for text


kmeht:
Having to write out the content of pages inside the React component itself is annoying, and often breaks the syntax highlighting. There should be a library out there that can convert Markdown to simple React components. This way, we should be able to write in simple text, and it won't ruin everything.

One hiccup would be: how do we do string interpolation like that, e.g. how do we insert our local variables into the string?

PR #595, opened 2017-05-26
uses markdown for text.

tdowds:
Related to #559

This PR is to introduce the functionality necessary to begin using markdown for some of our lengthier text in our components. To do this I began by introducing two dependencies (`markdown-loader` and `html-loader`) that would enable us to require markdown files as html strings.

Next, I introduced a new function in `huxley/www/js/utils/` that takes a markdown string and injects our variables into it.

Finally, I introduced a third dependency (`html-to-react`) that transforms this html string into a react component that we can place in our app's component tree.

Introducing three new dependencies isn't really ideal, but I wanted to push this up to start brainstorming approaches more if this proves insufficient. I've used AdvisorProfileView as an example of this approach.


package.json
```diff
@@ -9,8 +9,11 @@
     "fbjs": "0.8.3",
     "flux": "^2.1.1",
     "history": "^1.12.0",
+    "html-loader": "^0.4.5",
```
kmeht: This should be a dev dependency

package.json
```diff
@@ -9,8 +9,11 @@
...
+    "html-to-react": "^1.2.8",
     "jquery": "2.1.1",
     "js-cookie": "~2.0.3",
+    "markdown-loader": "^2.0.0",
```
kmeht: ditto

huxley/www/js/components/MarkdownTemplate.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * Copyright (c) 2011-2017 Berkeley Model United Nations. All rights reserved.
+ * Use of this source code is governed by a BSD License (see LICENSE).
+ */
+
+'use strict';
+
+var React = require('react');
```
kmeht: Replace all `var` with `const` or `let` as appropriate.

huxley/www/js/components/MarkdownTemplate.js
```diff
@@ -0,0 +1,46 @@
...
+      var regex = new RegExp("{{ " + variable + " }}", "g");
+      var value  = this.escapeHtml(this.props[variable]);
+      markdown = markdown.replace(regex, value);
+    }
+
+    return {__html: markdown};
+  },
+
+  escapeHtml: function(string) {
+    var entityMap = {
```
kmeht: Extract this to a variable in the module scope instead.

huxley/www/js/components/MarkdownTemplate.js
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright (c) 2011-2017 Berkeley Model United Nations. All rights reserved.
+ * Use of this source code is governed by a BSD License (see LICENSE).
+ */
+
+'use strict';
+
+var React = require('react');
+
+var MarkdownTemplate = React.createClass({
+  render: function() {
```
kmeht: Method syntax for new components:
```js
render() {
  //...
},
```

huxley/www/js/markdown/AdvisorProfileView.md
```diff
@@ -0,0 +1,9 @@
+## Profile
+
+Welcome, {{ first_name }}! We are very excited to see {{ school_name }} at BMUN {{ conference_session }} this year! Here, you can view and edit your registration information for the conference. Please note that fees are currently **estimates** based on the approximate delegation size given during registration. You can find more information on our fees [here](http://www.bmun.org/conference-fees/).
```
kmeht: May as well make these lowerCamelCase so that the props can be consistently named with the rest of the app, e.g. `firstName`, `schoolName`, etc.

huxley/www/js/components/AdvisorProfileView.js
```diff
@@ -20,6 +21,7 @@ var User = require('utils/User');
 var _handleChange = require('utils/_handleChange');
 
 require('css/Table.less');
+const AdvisorProfileViewMarkdown = require('markdown/AdvisorProfileView.md');
```
kmeht: nit: I'd place this in a folder called `text` instead of `markdown`, so we're not implying we're tied to a particular implementation. Similarly to how our LESS files are in `css`. I'd also call this variable `AdvisorProfileViewText`.

huxley/www/js/components/AdvisorProfileView.js
```diff
@@ -12,6 +12,7 @@ var InnerView = require('components/InnerView');
 var LogoutButton = require('components/LogoutButton');
 var ConferenceContext = require('components/ConferenceContext');
 var CurrentUserActions = require('actions/CurrentUserActions');
+var MarkdownTemplate = require('components/MarkdownTemplate');
```
kmeht: nit: Maybe `TextTemplate`, because the component doesn't actually know or care that the text is in markdown format; it only cares that it has the template variables in there.

PR #597, opened 2017-06-16
Code Cleanup on RegistrationView

jaketibbetts:
Took each section and turned it into it's own component. Hope it makes the code slightly more readable and easier to work with later on.


huxley/www/js/components/RegistrationView.js
```diff
@@ -724,4 +614,484 @@ const RegistrationPhoneInput = React.createClass({
   },
 });
 
-module.exports = RegistrationView;
+const RegistrationHeader = React.createClass({
```
m-j-mcdonald: Each component should be in its own file

huxley/www/js/components/RegistrationView.js
```diff
@@ -112,301 +116,187 @@ var RegistrationView = React.createClass({
 
   render: function() {
     var conference = this.context.conference;
+
+    //I feel like this (And a lot of others in this section) could 
+    //Be reduced with some kind of mapping,
+    //But there's no obvious way to map over Objects in JS so :/
+    var accountInformationErrors = {
```
m-j-mcdonald: If you want to map over objects, there are ways to make functions that can do it. But I don't see the explicit need to separate these into variables, the information should be tracked as part of the state

m-j-mcdonald: I'm open to discussing organizing the code like this though if you feel this improves readability

m-j-mcdonald: But I don't see the reason not to construct the objects as needed when they're passed to the child component

jaketibbetts: Constructing objects was really just to make the code clearer to read and organization slightly more manageable. But I agree that it has its setbacks and if there's a better way to do it, I'm here to learn.

And I figure if we add/subtract anything to any of the subcomponents in the future, it's slightly easier to modify when we pass them as props to subcomponents rather than to pass each individual piece of data as it's own prop. That's just my instinct though.

huxley/www/js/components/RegistrationView.js
jaketibbetts: This is actually a question that I have. Should we put each of these components into their own file even if we'll never use them outside of the RegistrationView?

m-j-mcdonald: Always put components into their own files; the code becomes impossible to read/maintain if there are a dozen components in one file

huxley/www/js/components/RegistrationView.js
m-j-mcdonald: For now at least I would leave them out and stick to the conventions we've been using; later on we can work out if we prefer a different structure

jaketibbetts: Okay sounds good!

huxley/www/js/components/RegistrationView.js
jaketibbetts: Thanks for answering that for me

m-j-mcdonald:
This looks good to me; what do you think @kmeht?

PR #598, opened 2017-07-05
updates to django 1.10.7 and DRF 3.6.3

tdowds:
- [x] update requirements and pass tests
- [x] update [middleware](https://docs.djangoproject.com/en/1.11/releases/1.10/#new-style-middleware)


huxley/core/middlewares.py
```diff
@@ -1,35 +1,46 @@
...
-        exc_type, exc_value, exc_traceback = sys.exc_info()
-        exc_traceback = ''.join(traceback.format_exception(exc_type, exc_value, exc_traceback))
-        log = json.dumps({
-              'message': exc_traceback,
-              'uri': request.path,
-              'status_code': 500,
-              'username': request.user.username})
-        logger.exception(log)
+  def __init__(self, get_response):
+    self.get_response = get_response
```
kmeht: Why do we have to do this?

huxley/core/middlewares.py
```diff
@@ -1,35 +1,46 @@
...
+
+  def __call__(self, request):
+    return self.get_response(request)
```
kmeht: And this?

huxley/core/middlewares.py
```diff
@@ -1,35 +1,46 @@
-# Copyright (c) 2011-2015 Berkeley Model United Nations. All rights reserved.
+# Copyright (c) 2011-2017 Berkeley Model United Nations. All rights reserved.
 # Use of this source code is governed by a BSD License (see LICENSE).
 
 import json, logging, sys, traceback
 
 
 class ExceptionLoggerMiddleware(object):
```
kmeht: This is probably a good opportunity to move this middleware file from `core` to `logging`, since this is all logging middleware. Also, let's follow the convention of naming it `middleware` singular. Remember to update the location in settings as well!

huxley/core/middlewares.py
tdowds: I think `__call__` is called by default (without it, you get `TypeError: 'ExceptionLoggerMiddleware' object is not callable`). And in it, it expects you to return the response object.

huxley/core/middlewares.py
tdowds: Middleware factories must accept a `get_response object`, and I think `__init__` is the means by which class based middleware factories accept it. [Link](https://docs.djangoproject.com/en/1.11/topics/http/middleware/#init-get-response)

Issue #600, opened 2017-07-15
Use prettier for code formatting


kmeht:
Similarly to how we use an autoformatter for Python code, let's also use [Prettier](https://github.com/prettier/prettier) for JS code so we don't have to worry about it.

Issue #550, opened 2016-12-29
Autoformatter doesn't handle deleted files


kmeht:
@tdowds ran into this issue: if you delete a file, the `fab format` throws.

PR #603, opened 2017-07-26
Introduces prettier and fixes autoformatter

tdowds:
This PR addresses #600 and #550 

To introduce prettier, I created a fab command that should be our only way to invoke prettier. Therefore, whether calling prettier on a standalone file from the command line or integrating it into our fabric workflow we maintain consistency with the options we specify in the command.

As for the options themselves, explicitly this is how the current configuration will format for us:
- line length of 80 characters
- 2 spaces per indentation-level
- indents with spaces rather than tabs
- adds a semi-colon at the end of every statement
- double quotes
- trailing commas wherever possible
- print spaces between brackets in object literals
- Put the `>` of a multi-line JSX element alone on the next line instead of at the end of the last line

Reference source for options is [here](https://github.com/prettier/prettier#options). If there is anything we need to change here, I'll push up in the next diff.

Additionally, I believe the autoformatter was throwing on a deleted file because we were formatting a file that doesn't exist anymore and then trying to commit on it. I haven't tested the fix I introduced yet.

### DIFF
--- a/fabfile/__init__.py
+++ b/fabfile/__init__.py
@@ -86,9 +86,10 @@ def format_python(py_diff_list):
 def format_js(source):
     '''Formats the javascript files in source using prettier. The argument
        source is a string where each file path is separated by a space.'''
-    options = '--trailing-comma all --jsx-bracket-same-line --no-bracket-spacing'
-    local('./node_modules/.bin/prettier ' + options + ' --debug-check ' + source)
-    local('./node_modules/.bin/prettier ' + options + ' --write ' + source)
+    if len(source) > 0:
+        options = '--trailing-comma all --jsx-bracket-same-line --no-bracket-spacing'
+        local('./node_modules/.bin/prettier ' + options + ' --debug-check ' + source)
+        local('./node_modules/.bin/prettier ' + options + ' --write ' + source)
 
 
 @task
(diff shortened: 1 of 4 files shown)


fabfile/__init__.py
```diff
@@ -68,10 +80,15 @@ def format():
...
 
-    if len(local('git status --porcelain', capture=True)) > 0:
-        local('git commit -am "Ran autoformatter"')
+
+@task
+def format_js(source):
+    '''Formats the javascript files in source using prettier. The argument
+       source is a string where each file path is separated by a space.'''
+    if len(source) > 0:
+        options = '--trailing-comma all'
```
kmeht: There's also an option for JSX to put the last `>` on the line of the last prop, let's use that too.

fabfile/__init__.py
```diff
@@ -55,10 +55,22 @@ def lint():
...
 @task
 def format():
-    '''Format and commit python files committed on the current feature branch'''
-    diff_list = git.diff_name_only()
+    '''Format files committed on the current feature branch'''
+    diff_list = git.diff_name_only(options='--diff-filter=d')
     py_diff_list = [pyfile for pyfile in diff_list if pyfile.endswith('.py')]
+    format_python(py_diff_list)
+
+    js_source = ' '.join(
```
kmeht: Only if there are actually any JS files, right?

tdowds: yeah, in that case `js_source` will be an empty string

Issue #531, opened 2016-10-21
Test RegistrationView component behavior


tdowds:
It gets pretty boring pretty fast running when for every change we make to our application we run through a mock registration to make sure our application still works. Instead what we should do is test the behavior of our UI components.

Ideally what this would look like is I can run a test that will go through the registration process for me and make sure that whatever behavior we expect to see in `RegistrationView` occurs when we specify an interaction with it.

Things we should test for are:

1) Display of front end validation
2) UI changes due to back end validation 
3) UI changes due to successful registration

and probably many, many more.

For this particular task we should limit our scope to just the RegistrationView, but ideally we would do this for each of our front end views.

Some resources that might help with this include:

1) Googling "headless browser testing"
2) Which will probably lead you to [PhantomJS](http://phantomjs.org/)
3) And some other services, but I've used PhantomJS before and it seems pretty common

If anyone else has recommendations on what we should use feel free to edit this issue description :)

PR #605, opened 2017-07-30
Bug fixes on RegistrationView

tdowds:
This is a PR to catch some of the bugs on our refactored `RegistrationView`

- [x] doesn't display error messages from the server
- [x] password confirmation error
- [x] checkboxes in the special committee preferences section wasn't showing up

#531 might help in testing our components more thoroughly when making changes like this in the future


m-j-mcdonald:
These are all good catches. I'll assign someone to #531 to help with this.

Issue #331, opened 2014-09-22
Split out a Registration model from School


kmeht:
Currently, the School model holds two categories of data: the actual school information (name, address, etc.), but also information specific to its registration for a particular session of the conference (number of delegates, country preferences, etc).

We should split this into two models: School, which contains the infrequently-changing school information, and Registration, which depends on School and Conference. This will allow us to build features like registration reactivation, where a school doesn't have to put in their basic information again; they can just update their registration information and re-register immediately!

This will be a pretty non-trivial task, involving at least three migrations, but I think it will be well worth it. I believe that Registration will be the central model of the application, rather than School.

kmeht:
@nathanielparke, note that we'll probably need to add a Conference model before handling this task as well. We can probably make that a separate task.

tdowds:
@m-j-mcdonald pretty sure this is all done now. Let me know if you encounter any bugs from this when getting ready to push to prod or around the time of registration opening up. Also, these PRs would be good for younger members as a learning reference.

Issue #109, opened 2013-09-18
Improve our user emails with templates.


kmeht:
Currently, we directly email the user with a short string. Instead, we can offer longer, richer emails with templates. It requires a bit of extra code, so it'd best to abstract this functionality into a utility.

We can then have a separate template directory for our emails, and then send them with whatever utility we create to wrap them.

StackOverflow had a really informative answer on how to achieve this: http://stackoverflow.com/questions/2809547/creating-email-templates-with-django.

Issue #48, opened 2013-06-20
Registration Reactivation


kmeht:
An advisor registering for the next BMUN session shouldn't have to input all this information again. Most of it will remain the same. Let's add a one-click reactivation feature that allows advisors to re-register with the conference after updating their profile information.

This will require some logic changes, because currently we assume that any existing advisor account is registered for the conference. The `is_active` property of Django User models will come in handy here.

We'll also have to create a `Registration` many-to-many association between schools (or advisors, or whatever) and Conferences. This way, advisors can register for various conferences.

PR #606, opened 2017-08-01
Introduce Registration model

tdowds:
This PR is to split out a Registration model from the School model as defined in #331. Some things that need to be done:

- [ ] agree on what fields to leave in School and what fields to place in Registration
- [ ] add Registration to admin panel
- [ ] update class methods in Registration model
- [ ] update api (views, serializers, validators, urls, permissions, etc) for School and Registration
- [ ] update back end tests
- [ ] update front end data flow to reflect changes in models


tdowds:
This first diff, is to gain a consensus on what fields should go into which models. Any opinions would be much appreciated.

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+class Registration(models.Model):
+    CONTACT_TYPE_OPTIONS = ((ContactType.FACULTY, 'Faculty'),
+                            (ContactType.STUDENT, 'Student'), )
+
+    GENDER_OPTIONS = ((ContactGender.MALE, 'Male'),
+                      (ContactGender.FEMALE, 'Female'),
+                      (ContactGender.OTHER, 'Other'),
+                      (ContactGender.UNSPECIFIED, 'Unspecified'), )
+
+    registered = models.DateTimeField(auto_now_add=True)
```
kmeht: Maybe `date_registered` or `registered_at` or `created_at`?

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+    secondary_name = models.CharField(max_length=128, blank=True)
+    secondary_gender = models.PositiveSmallIntegerField(
+        choices=GENDER_OPTIONS, blank=True, default=ContactGender.UNSPECIFIED)
+    secondary_email = models.EmailField(blank=True)
+    secondary_phone = models.CharField(max_length=32, blank=True)
+    secondary_type = models.PositiveSmallIntegerField(
+        choices=CONTACT_TYPE_OPTIONS, blank=True, default=ContactType.FACULTY)
+
+    times_attended = models.PositiveSmallIntegerField(default=0)
+    beginner_delegates = models.PositiveSmallIntegerField()
```
kmeht: `num_beginner_delegates`, ditto for the other four. Maybe a little verbose, but worth it to be explicit? What do you think?

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+    intermediate_delegates = models.PositiveSmallIntegerField()
+    advanced_delegates = models.PositiveSmallIntegerField()
+    spanish_speaking_delegates = models.PositiveSmallIntegerField()
+    chinese_speaking_delegates = models.PositiveSmallIntegerField()
+
+    countrypreferences = models.ManyToManyField(
```
kmeht: `country_preferences`

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+        Country, through='CountryPreference')
+    committeepreferences = models.ManyToManyField(
```
kmeht: `committee_preferences`

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+                    'If you have any tech related questions, please email tech@bmun.org. '
+                    'For all other questions, please email info@bmun.org.\n\n'
+                    'Thank you for using Huxley!' % conference.session,
+                    'no-reply@bmun.org', [school.primary_email],
+                    fail_silently=False)
+            else:
+                registration_fee = conference.registration_fee
+                delegate_fee = conference.delegate_fee
+                send_mail(
+                    'BMUN %d Registration Confirmation' % conference.session,
```
kmeht: Gawd damn; is anyone actually working on #109?

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+                    fail_silently=False)
+
+    def __unicode__(self):
+        return self.name
+
+    class Meta:
+        db_table = u'school'
+
+
+pre_save.connect(School.update_fees, sender=School)
```
kmeht: These should all be `Registration`, no?

tdowds: Yeah I think. I was gonna go in and fix all this kind of stuff after we think both models are properly defined

m-j-mcdonald:
@tdowds This was an issue I'd thought about giving to some new members, along with #48, once they had gone through their initial issues. Since you've got the ball rolling do you want the help or do you want to keep at it by yourself?

The PR looks good so far! The only thing we might want to change is to have times_attended as part of the school so that it can be easily incremented at each new registration once #48 is in place.

huxley/core/models.py
```diff
@@ -444,3 +270,195 @@ def save(self, *args, **kwargs):
...
+
+    modified_at = models.DateTimeField(default=timezone.now)
+
+    def balance(self):
+        return self.fees_owed - self.fees_paid
+
+    def update_country_preferences(self, country_ids):
+        '''Given a list of country IDs, first dedupe and filter out 0s, then
+        clear the existing country preferences and construct new ones.'''
+        seen = set()
```
m-j-mcdonald: I think we can remove the handling for duplicates here if we want to; the frontend has had a feature for awhile to prevent selecting the same country twice.

tdowds:
@m-j-mcdonald I wouldn't mind helping out a younger member with this issue - it would be an awesome learning experience for him or her. However, I'm concerned that the time spent working on this issue might spill over into when registration goes live. In my experience, working with younger members slows the development process. Additionally, I'm not entirely sure how much time I'll have to devote to this once I start work in two weeks. 

If it does spill over we'd have to migrate over live data to fit our new models and work out any bugs live that our tests don't catch. In this case, I think the backwards compatible approach that @kmeht suggested would be the way we have to go. 

If you guys think we can finish this before registration opens (with maybe sometime to catch some bugs), or if merging this while registration is live isn't an issue I'm all for doing this with our without a younger member to work on this with. I just want us all to be on the same page about the commitment required to get this done given the timing.

kmeht:
@m-j-mcdonald From my experience, this is above the level of a newer developer. You would be surprised at how long the ramp-up time is for someone brand new to building software.

kmeht:
(That said, it would be extremely helpful for newer members to follow-along on this PR as a learning experience, but I think @tdowds going solo is the best bet at getting this done before registration).

tdowds:
This should define the Registration model properly. If we want to move some fields around, now would be a good time to mention it. I'll next go in and add all the functionality around the model to get it working in the app before removing that functionality from the School model.

kmeht:
Getting the app working with Registration before deleting the fields from School is a great idea. Also, look into data migrations, which can help you move all of the data from School into Registration before you write the migration to delete School.

This way, we have the option to build Registration Reactivation as well, and we can reuse the existing data from all the folks who registered for BMUN 2017.

It's worth noting that all of this doesn't have to be one PR. For example, you could land this (once it's ready), and introduce the API, etc. in followup PRs.

huxley/core/models.py
```diff
@@ -272,6 +272,131 @@ class Meta:
...
+                     int(delegate_fee)),
+                    'no-reply@bmun.org', [registration.school.primary_email],
+                    fail_silently=False)
+
+    def __unicode__(self):
+        return self.school.name + ' - ' + self.conference.session
+
+    class Meta:
+        db_table = u'registration'
+        unique_together = ('conference', 'school')
```
kmeht: Nice constraint, love it.

huxley/core/models.py
```diff
@@ -383,10 +509,13 @@ class Meta:
         db_table = u'assignment'
         unique_together = ('committee', 'country')
 
+
 pre_save.connect(Assignment.update_assignment, sender=Assignment)
 
+
 class CountryPreference(models.Model):
     school = models.ForeignKey(School)
```
kmeht: We can keep the school for now, but we should plan to access the school via the registration. Otherwise the information is redundant. This can be done in followup PRs; I'm wary of one PR getting too big.

huxley/core/models.py
```diff
@@ -272,6 +272,131 @@ class Meta:
...
+        unique_together = ('conference', 'school')
+
+
+pre_save.connect(Registration.update_fees, sender=Registration)
+pre_save.connect(Registration.update_waitlist, sender=Registration)
+post_save.connect(Registration.email_comments, sender=Registration)
+post_save.connect(Registration.email_confirmation, sender=Registration)
+
+
 class Assignment(models.Model):
```
kmeht: Assignment should also have a `registration` field, because it is per-conference data.

tdowds: Should we plan to access the school via the registration here as well like we plan to do in CountryPreference?

tdowds:
If this all looks good, I'll merge after I fix whatever suggestions @m-j-mcdonald might have.

huxley/core/models.py
kmeht: Yes, similarly to CountryPreference, the school field will be removed.

m-j-mcdonald:
This all looks good to me; I'd say go ahead and merge.

Issue #580, opened 2017-03-19
Remove delegate feedback state from Delegate model


tdowds:
Given @kmeht's comments in [this pr](https://github.com/bmun/huxley/pull/561), we should rethink how we're storing our feedback data.

tdowds:
Don't know where you guys are at with this, but might not be a bad idea to post here whatever designs you have in mind for storing this information before hopping into the code.

m-j-mcdonald:
I've been working with Jake & Jake on how to structure it. It's going to create a feedback model that will belong to an assignment (assignments will have a one-to-one field with it). The alternative was having the feedback model have an assignment but there weren't any strong advantages to that and this seemed more intuitive

PR #607, opened 2017-08-01
Clean up ChairView code

m-j-mcdonald:
This makes a series of general improvements to the chair app code, partly in anticipation of #580.

Some things changed include:
- Map the relevant fields by assignment, making them independent of the delegate models
- Create new objects at state updates instead of abusing the state
- Clear the timeouts on the animations to avoid setting the state after unmount

I do think the sort calls are less elegant here but I felt it was better to place them where they would be called once instead of at every render.

### DIFF
--- a/huxley/www/js/components/ChairAttendanceView.js
+++ b/huxley/www/js/components/ChairAttendanceView.js
@@ -79,7 +79,7 @@ var ChairAttendanceView = React.createClass({
       }
       this.setState({
         delegates: delegates,
-        attendance: Object.assign({}, attendance, update),
+        attendance: {...attendance, ...update}),
       });
     });
 
@@ -176,11 +176,11 @@ var ChairAttendanceView = React.createClass({
   _handleAttendanceChange(field, assignmentID, event) {
     var attendanceMap = this.state.attendance;
     var oldAttendance = attendanceMap[assignmentID];
-    var newAttendance = Object.assign({}, oldAttendance);
+    var newAttendance = {...oldAttendance};
     newAttendance[field] = !newAttendance[field];
     var update = {};
     update[assignmentID] = newAttendance;
-    this.setState({attendance: Object.assign({}, attendanceMap, update)});
+    this.setState({attendance: {...attendanceMap, ...update}});
   },
 
   _handleSaveAttendance(event) {
@@ -198,7 +198,7 @@ var ChairAttendanceView = React.createClass({
                       session_two: attendance["session_two"],
                       session_three: attendance["session_three"],
                       session_four: attendance["session_four"]};
-        toSave.push(Object.assign({}, delegate, update))
+        toSave.push({...delegate, ...update})
       }
     }
     DelegateActions.updateCommitteeDelegates(committee, toSave, this._handleSuccess, this._handleError);
--- a/huxley/www/js/components/ChairAttendanceView.js
+++ b/huxley/www/js/components/ChairAttendanceView.js
@@ -30,7 +30,7 @@ var ChairAttendanceView = React.createClass({
     var assignments = AssignmentStore.getCommitteeAssignments(user.committee);
     var countries = CountryStore.getCountries();
     var delegates = DelegateStore.getCommitteeDelegates(user.committee);
-    var attendance = {};
+    var attendance = this._mapAttendance(delegates);
     if (assignments.length && Object.keys(countries).length) {
       assignments.sort(
         (a1, a2) =>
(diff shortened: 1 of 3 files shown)


huxley/www/js/components/DelegationAttendanceRow.js
```diff
@@ -11,8 +11,17 @@ var DelegationAttendanceRow = React.createClass({
   propTypes: {
     onChange: React.PropTypes.func,
     countryName: React.PropTypes.string,
-    countryID: React.PropTypes.string,
-    delegates: React.PropTypes.array,
+    assignmentID: React.PropTypes.number,
+    attendance: React.PropTypes.object,
+  },
+
+  shouldComponentUpdate(nextProps, nextState) {
```
kmeht: This is a dangerous `shouldComponentUpdate`, because it won't re-render the component if the `assignmentID` changes. In theory, it never should, but this component cannot make any guarantees as to whether it'll be used correctly.

huxley/www/js/components/DelegationAttendanceRow.js
```diff
@@ -26,7 +35,7 @@ var DelegationAttendanceRow = React.createClass({
             <input
               className="choice"
               type="checkbox"
-              checked={this.props.delegates[0].voting}
+              checked={this.props.attendance['voting']}
```
kmeht: Why switch to the indexing syntax from the property access syntax?

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -188,8 +192,9 @@ var ChairSummaryView = React.createClass({
   _handleSummaryChange(assignment, event) {
     var newSummary = event.target.value;
     var summaries = this.state.summaries;
-    summaries[assignment.id] = newSummary;
-    this.setState({summaries: summaries});
+    var update = {};
+    update[assignment.id] = newSummary;
+    this.setState({summaries: {...summaries, ...update}});
```
kmeht: Don't need an update object here, if you used computed property syntax (which I'm pretty sure we have the transform for):
```js
this.setState({
  summaries: {
    ...this.state.summaries,
    [assignment.id]: event.target.value,
  },
});
```

huxley/www/js/components/ChairSummaryView.js
```diff
@@ -55,20 +62,17 @@ var ChairSummaryView = React.createClass({
...
+      var summaries = this.state.summaries;
+      var update = {};
       for (var delegate of delegates) {
-        summaries[delegate.assignment] = delegate.summary;
+        update[delegate.assignment] = delegate.summary;
       }
       this.setState({
         delegates: delegates,
-        summaries: summaries,
+        summaries: {...summaries, ...update},
```
kmeht: Love it!

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -27,13 +27,35 @@ var ChairAttendanceView = React.createClass({
...
+    var attendance = {};
+    if (assignments.length && Object.keys(countries).length) {
+      assignments.sort(
+        (a1, a2) =>
+          countries[a1.country].name < countries[a2.country].name ? -1 : 1,
+      );
+    }
+
+    for (var delegate of delegates) {
+      var delegateAttendance = {};
```
kmeht: Why not use an object literal?
```js
attendance[delegate.assignment] = {
  voting: delegate.voting,
  session_one: delegate.session_one,
  //...
};
```

huxley/www/js/components/ChairAttendanceView.js
```diff
@@ -46,23 +68,51 @@ var ChairAttendanceView = React.createClass({
 
   componentDidMount() {
     var user = CurrentUserStore.getCurrentUser();
-    this._mapAssignments();
+    var attendance = this.state.attendance;
+
     this._delegatesToken = DelegateStore.addListener(() => {
+      var delegates = DelegateStore.getCommitteeDelegates(user.committee);
+      var update = {};
+      for (var delegate of delegates) {
```
kmeht: This is duplicated code from above; extract into a function that takes an array of delegates and converts it to the attendance map.

huxley/www/js/components/DelegationAttendanceRow.js
m-j-mcdonald: I added a check on assignmentID; that was an oversight not to include form the start. We should have some sort of check on should it update though since for a large committee there may be hundreds of these rows and I think it would be inefficient to re-render every time one is updated.

kmeht:
We should start separating our main View components into Container and View pairs, so that all the state management logic doesn't litter the display logic.

PR #608, opened 2017-08-03
Introduces Registration API

tdowds:
This PR is to introduce the API for plugging into our Registration model ( #331 )  . I would imagine we would want to do the following:

1. POST data
2. GET data based off school and conference session
3. Update and destroy

Let me know your thoughts!

### DIFF
--- a/huxley/api/permissions.py
+++ b/huxley/api/permissions.py
@@ -6,7 +6,7 @@ import json
 from django.http import QueryDict
 from rest_framework import permissions
 
-from huxley.core.models import Assignment, Delegate
+from huxley.core.models import Assignment, Delegate, Registration
 
 
 class IsSuperuserOrReadOnly(permissions.BasePermission):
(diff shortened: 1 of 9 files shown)


tdowds:
This first diff enables us to POST data to our model. Assuming this is good to go, I'll next focus on getting, updating and destroying.

m-j-mcdonald:
I think it would be worth defining the permissions class with this PR

tdowds:
Added the permissions. I think there's room for discussion on the permissions. For example, do we want to allow anyone to post registration information? For now though, what has been pushed up should leave us with the most flexibility.

I added in the capability to query by `school_id` and `conference_id`. I can imagine that will be useful later on.

For now, I don't see a need to add a `RegistrationDetail` view. For the sake of time it might be worth putting that off until we need that endpoint. But I can throw it up if necessary.

Also @kmeht we need to put together a timeline today of when we want to finish this entire feature for external. Let's sync up when you get the chance.

PR #610, opened 2017-08-07
Created data migration for Registration model 

tdowds:
This migrates the registration data from school instances in the current db to new registration instances defined by our Registration model ( #331 ). 

Django data migration [docs](https://docs.djangoproject.com/en/1.10/topics/migrations/#data-migrations) for reference.


m-j-mcdonald:
Looks good to me!

PR #612, opened 2017-08-10
Delegate Accounts Backend

m-j-mcdonald:
This is the first PR to introduce delegate accounts into Huxley. This handles the backend structure of the accounts; the next PR will introduce the frontend views and then the final PR will introduce the code to create the accounts.

My objective for introducing these is two fold: the primary is that I intend to incorporate position paper grading into Huxley and before that begins delegates will need a portal to access Huxley. But also, it gives us a better way to route the feedback from conference and provide a general interface for interacting with delegates.


m-j-mcdonald:
Also, apologies for all the changes made by the auto formatter. It made the diff quite a bit longer.

tdowds:
This looks very solid and very exciting. I apologize in advance for the gnarly merge conflicts that will probably have to be resolved, but once you get those done, this will probably be good to go.

m-j-mcdonald:
Are there any objections to me merging this? There are two more PRs to complete the issue and there might be some debate over the last one, so I'd like to get them up as soon as possible.

tdowds:
Sorry for being late to the game here - starting work has been hectic. But this is some great work. Looks good to me. Excited to see what comes next

PR #616, opened 2017-08-22
Enable registration with Registration model

tdowds:
This enables us to register new users and schools using our Registration model ( #331 ). Things that need to be done here include:

1. Successfully post user, school and registration information to our database
2. Return errors to the client upon a bad request
3. Render errors on the client
4. Write some tests


tdowds:
This first diff is for 1 and 2. Let me know your thoughts if you think there is a better way to approach creating a User instance, School instance, and Registration instance in one request.

tdowds:
4

tdowds:
3

huxley/www/js/components/registration/RegistrationPhoneInput.js
```diff
@@ -18,10 +18,6 @@ const RegistrationPhoneInput = React.createClass({
     value: React.PropTypes.string,
   },
 
-  shouldComponentUpdate: function(nextProps, nextState) {
```
tdowds: removed this because error messages for primary_phone weren't showing

m-j-mcdonald:
I think this all looks good, except I'm not sure we should have registration creation in a mixin; it seems specific enough to the register view class that the methods could just belong to that

tdowds:
Threw in a test to enforce the structure of the response on a bad request. Also I incorporated @m-j-mcdonald's suggestion - good catch

Issue #614, opened 2017-08-13
Delegate Accounts


m-j-mcdonald:
It's time we put delegate accounts into Huxley. The objective is two fold: the primary aspect being that I intend to incorporate position paper grading into Huxley, and before that begins delegates will need a portal to access Huxley. But also, it gives us a better way to route feedback from conference and provide a general interface for interaction with delegates.

Here is a general overview to track the progress for these:
- [x] Update the backend model and api
- [x] Add initial frontend views as well as stores, actions, and routing
- [x] Create accounts when assignments are finalized
- [x] Delete accounts with delegates

kmeht:
@m-j-mcdonald I imagine deletion is also a part of this, yeah? Since advisors can not only add but delete delegates, we should delete the accounts as well.

m-j-mcdonald:
I don't think deletion is a huge concern, since advisors can't remove delegates once they have finalized their rosters (which happens before they finalize their assignments). That's part of why I designed he account creation to happen when assignments are finalized.

PR #617, opened 2017-08-23
Delegate Accounts Frontend

m-j-mcdonald:
This handles part 2 of issue #614. The profile is only a skeleton for right now, but it will be easy to build off of as we develop features.


tdowds:
This is super exciting. Nice work here. Something to consider is perhaps instead of having a `SchoolStore` we can query the school in our login view logic and pass it up in the Response. Might be worth considering doing that for the delegate's committee and assignment as well. I haven't fully considered the pros and cons of doing that and for what you're doing now, but if you've already fleshed that out would love to hear it or work on it with you.

tdowds:
Also, be sure to add a test file for the school store

m-j-mcdonald:
That was a good catch; fetching the models with the user definitely reduces the complexity of the code and the number of queries on the server.

m-j-mcdonald:
What are your thoughts on using the nestedserializers? I think it improves how we query the database but for post requests it could potentially complicate things.

tdowds:
I like where you're heading with it. I think it's a good idea to nest  a DelegateSerializer on the User Serializer. When a delegate logins we'll easily pass up his/her school and delegate information. When an advisor (or chair) logs in we'll probably have a delegate field with an empty object as a value, but I don't think that should cause any problems (except maybe in `updateUser` in `CurrentUserStore`). As for the equivalent POST request, it might make things less complicated by creating a `CreateDelegateUserSerializer` class that doesn't include a nested `SchoolSerializer` but does include a nested `DelegateSerializer` (not a `DelegateNestedSerializer`), that way you can pass in the advisor's school id instead of the whole school object (same for the assignment) when posting  (if that's what you mean by complicated).

I don't think an `AssignmentNestedSerializer` is totally necessary. We can grab those objects out of the `CommitteeStore` and `CountryStore` once we have their ids on the assignment object. Not as convenient, but it simplifies our payload coming back up (not totally sure if there will be issues down the road because we have a more complex payload (maybe issues with PUT/PATCH?)).

By the same reasoning, I'm a little uncertain of the `DelegateNestedSerializer` - will certainly make querying more convenient, but not sure how difficult it is to manage the complexity of a nested payload.

I do think creating a separate `CreateDelegateUserSerializer` will resolve complications with POST. As for the nested serializers, I'd say go for it, and let's see how it plays out. We have a clear alternative (falling back on the stores) if we run into something we can't handle, and we'll probably learn something along the way :p 

Let me know if this is unclear, or if you want to flesh this out more over messenger. Great work so far.

m-j-mcdonald:
Would you object if I push this up? I know I've let it sit for awhile while I was dealing with registration, which is my bad. I think that this will work fine for all our use cases; for position papers they'll probably be their own model and we can put in an api endpoint for them pretty easily.

tdowds:
Sounds good to me then. And no worries about letting it sit - one thing at time

PR #618, opened 2017-08-31
Front End Update for Registration Model

tdowds:
This PR is to update our front end code to work with our Registration model ( #331 ). Some things to do in this PR:

1. Get the Registration object and populate the correct fields
2. Update the Registration object when finalizing assignments
3. Write and update tests


tdowds:
1

tdowds:
2

tdowds:
3

m-j-mcdonald:
This looks good to me! The only thing I think might worth revisiting is the structure of the registrationstore; it theoretically could end up in a situation where it allows a user access to a registration they don't have permission to view (if a previous user accesses a registration and then somehow the new user queries that schoolID & conferenceID). The scenario's far fetched and I don't think it's an issue, but we could consider something like what we do in delegate store.

Issue #622, opened 2017-09-11
Registration Validation Bug


m-j-mcdonald:
During registration, calling is_valid on a serializer will first return the errors from field-specific validators, if there are any, and then only when there are no errors from those return the errors from validate(). 

This means in api/view/registers the initial call to is_registration_valid won't catch the errors from validate (since there's no school), then we'll save the user & school but the registration fails since now it sees the errors from validate(). This results in a scenario where the user & school are saved but not the registration, preventing another registration for those user/school names because they technically already exist.

I'm pushing up a temporary fix for registration (removing the validate method that's causing the issue) but we need to find a better way to handle this.

m-j-mcdonald:
To reproduce (with the validate method in the registration serializer uncommented):

Fill out a registration, but have the number of spanish or chinese speaking delegates exceed the total delegates you put in elsewhere and hit register. The registration should fail. (Notice the error message does not propagate up; this is a direct symptom of the bug). Then fix the issue by adjusting the number of spanish/chinese speaking delegates down. Try to register again; it'll say that username/school already exist.

PR #624, opened 2017-09-12
Fixes registration validation bug

tdowds:
Fix for #622 

The issue appears to be that the validation method of a DRF serializer doesn't get called when a required field is empty. With the way `RegisterView` works, it assumes that the school field of the registration data of the request will be empty until the school is created on the User. Unfortunately, this means that the validation method of `RegistrationSerializer` will never be called because the school field will always be empty based off our assumption. The fix for this proposed here is to temporarily set the school field to read only, so that way it doesn't do any validation on the school field of the Registration object. 

Here are some relevant links from DRF's source code for reference:
[is_valid](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L222)
[run_validation](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L422)
[to_internal_value](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L442)

### DIFF
--- a/huxley/api/serializers/registration.py
+++ b/huxley/api/serializers/registration.py
@@ -50,30 +50,36 @@ class RegistrationSerializer(serializers.ModelSerializer):
             'registration_comments': {'required': False}
         }
 
-    ## This causes an issue during registration; see issue #622
-    # def validate(self, data):
-    #     invalid_fields = {}
 
-    #     num_beginner_delegates = data.get('num_beginner_delegates')
-    #     num_intermediate_delegates = data.get('num_intermediate_delegates')
-    #     num_advanced_delegates = data.get('num_advanced_delegates')
-    #     num_spanish_speaking_delegates = data.get(
-    #         'num_spanish_speaking_delegates')
-    #     num_chinese_speaking_delegates = data.get(
-    #         'num_chinese_speaking_delegates')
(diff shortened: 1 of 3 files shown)


tdowds:
Removed the hack in the last diff, and instead use `transaction.atomic` to create our model instances. This way if any validation fails, no model instance will be created. If there are no validation errors, everything will run just fine. The only down side is that now if a user submits invalid inputs for User/School and Registration, the errors will come back for User/School but not for Registration until User/School gets fixed.

tdowds:
It now will show all the validation errors on a given input instead of just showing the errors of User/School even though User/School and Registration have errors.

tdowds:
I think this is good to go. I updated the test file a little bit to try and do a better job of catching any sneaky bugs like the one that motivated this PR. What is being tested is:

1. General success and asserting the structure of that response
2. Given the order in which the model instances are created, a User and School model instance cannot be created while a Registration model instance is not
3. In a response to a bad request, any errors for the school object are not overwritten by the errors from the registration object saying that a school is required.
4. Ensuring that all errors are returned and not just one model's errors (e.g. User's) on a bad request

These are some of the requirements and quirky bugs I've identified and tried to test for. If you can think of more then we should test for them, else I say we ship.

PR #628, opened 2017-09-16
More Registration Stuff

m-j-mcdonald:


kmeht:
@m-j-mcdonald why commented out instead of removed completely? Or is it just that the conference didn't have a Chinese committee this year, but may have one next year? Perhaps that means bilingual committees should be a first-class feature of the application. But that's certainly lower-pri than making sure core functionality works.

m-j-mcdonald:
The main reason was to avoid a substantial refactoring of the codebase given we may want to add it right back in next year. I admit I rushed this out a bit since it was a last minute request from External, but I think we should wait until we know what next year's bilingual situation will look like before making the call on cutting this out or making it a first class feature.

PR #630, opened 2017-09-23
Fixes Assignment Upload

### LINKED ISSUE #627: Assignment admin needs to be modified to use registration model
core/admin/assignment.py was never updated to reflect the registration model; just putting this as a reminder we need to do that before External tries to upload assignments

m-j-mcdonald:
This fixes #627 by modifying the assignment update function to account for the registration model, as well as fixing some other bugs I noticed:
1. We pass strings as a signal that a model does not exist, but we were checking model ids before checking that the objects were not strings
2. I can't think of any good reason to remove the remainder of assignment_dict unless we explicitly want to remove all old assignments, something External very emphatically does not want to happen. By taking that out now the uploaded csv just updates the set of assignments without replacing them.

### DIFF
--- a/huxley/core/models.py
+++ b/huxley/core/models.py
@@ -321,17 +321,6 @@ class Assignment(models.Model):
             deletions.append(assignment_data['id'])
 
         for committee, country, school, rejected in new_assignments:
-            key = (committee.id, country.id)
-            if key in assigned:
-                # Make sure that the same committee/country pair is not being
-                # given to more than one school in the upload
-                committee = str(committee.name)
-                country = str(country.name)
-                failed_assignments.append(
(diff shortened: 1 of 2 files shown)


kmeht:
Old assignments were deleted in case certain assignments ceased to exist,
e.g. what if USA were no longer in DISC? That assignment object should be
deleted.

Of course, it’s possible that the mental model for assignment upload has
been changed since I wrote that code, so just take this context into
consideration and continue as you were :)

m-j-mcdonald:
Okay, that makes sense. Now I think the mindset is more along the lines that those deletions can be done separately, while additions might be done in batches where a spreadsheet does not necessarily contain the assignments that need to be kept.

PR #631, opened 2017-09-28
Delegate Account Creation

m-j-mcdonald:
The final part of #614. In addition to account creation, this makes the delegate's email a required field (so that the delegate can receive an email with account login information).


m-j-mcdonald:
Does anyone have some ideas for good unit tests on this? I have a few but I want to be confident that the accounts work as intended.

huxley/www/js/components/DelegateSelect.js
```diff
@@ -18,8 +18,7 @@ var DelegateSelect = React.createClass({
     return (
       <select
         onChange={this.props.onChange}
-        value={this.props.selectedDelegateID}
-        disabled>
```
kmeht: Could you remind me why this was disabled in the first place?

huxley/api/serializers/delegate.py
```diff
@@ -25,6 +31,32 @@ class Meta:
...
+            names = instance.name.split(' ')
+            username = names[0] + '_' + str(instance.id)
+            password = BaseUserManager().make_random_password(10)
+            user = User.objects.create(delegate=instance, username=username,\
+                                       user_type=User.TYPE_DELEGATE,\
+                                       last_login=datetime.now(),
+                                       first_name=names[0], last_name=names[-1],
+                                       email=instance.email)
+            user.set_password(password)
+            user.save()
```
kmeht: We're hitting the DB twice here: once when we call `create`, and once when we call`save`. We should only do it once (right here). Instead of using the manager's `create` method, you should be able to just construct a User object. If there's a particular reason we're saving twice, feel free to ignore.

huxley/api/serializers/delegate.py
```diff
@@ -25,6 +31,32 @@ class Meta:
...
+            send_mail('A new account has been created for {0}!\n'.format(instance.name),
```
kmeht: We should really use templates for emails. We have #109 open for that, if you wanna assign that to a new developer.

huxley/api/serializers/delegate.py
```diff
@@ -25,6 +31,32 @@ class Meta:
...
+        if ('assignment' in validated_data and
+                validated_data['assignment'] != None and
+                not instance.assignment):
+
+            names = instance.name.split(' ')
+            username = names[0] + '_' + str(instance.id)
+            password = BaseUserManager().make_random_password(10)
+            user = User.objects.create(delegate=instance, username=username,\
+                                       user_type=User.TYPE_DELEGATE,\
+                                       last_login=datetime.now(),
```
kmeht: Is this a field we added? If it's default Django, wouldn't this be set by default?

huxley/api/serializers/delegate.py
```diff
@@ -25,6 +31,32 @@ class Meta:
...
+                                       first_name=names[0], last_name=names[-1],
+                                       email=instance.email)
+            user.set_password(password)
+            user.save()
+            send_mail('A new account has been created for {0}!\n'.format(instance.name),
+                      'Username: {0}\n'.format(username) \
+                      + 'Password: {0}\n'.format(password) \
```
kmeht: Yikes, plaintext passwords. But, since delegates aren't creating their own accounts, I don't see an immediate way around this, other than forcing them to change their password the first time they log in.

huxley/www/js/components/DelegateSelect.js
m-j-mcdonald: It was done here: [https://github.com/bmun/huxley/commit/64c84d2a1c24aa65543a221a96ca71fd9bea6089#diff-3bd1cda3536a01c6309345f7a007ca59](url), I assume to disable some functionality before conference. This ties into why I opened #632

huxley/api/serializers/delegate.py
m-j-mcdonald: I agree it isn't great, but the accounts themselves contain nothing sensitive and we need an efficient way to distribute the passwords.

huxley/api/serializers/delegate.py
m-j-mcdonald: I'll find someone this weekend to work on it

huxley/api/serializers/delegate.py
m-j-mcdonald: Good catch; I'll fix this.

huxley/api/serializers/delegate.py
m-j-mcdonald: It's something we set when creating accounts for advisors; I wasn't sure of the motivation there so I figured it wouldn't hurt to include it here as well.

PR #634, opened 2017-10-12
Delete Delegate Accounts

m-j-mcdonald:
This is to ensure accounts get deleted with their delegates (the last part of #614)


m-j-mcdonald:
I'm going to add more comprehensive tests now, but I also wanted to put all the functionality in place asap for some stress testing on stage

PR #635, opened 2017-10-19
Delegate Account Unit Tests and bug fixes

m-j-mcdonald:
This adds unit tests on creating/deleting the delegate accounts as well as catches some issues that manifested themselves in stage but not locally

### DIFF
--- a/huxley/api/tests/test_user.py
+++ b/huxley/api/tests/test_user.py
@@ -425,8 +425,7 @@ class DelegateUserCreateTestCase(tests.PartialUpdateAPITestCase):
         self.advisor = models.new_user(username='advisor', password='advisor')
         self.school = models.new_school(user=self.advisor)
         self.registration = models.new_registration(school=self.school)
-        self.assignment = models.new_assignment(
-            registration=self.registration)
+        self.assignment = models.new_assignment(registration=self.registration)
         self.delegate = models.new_delegate(
             school=self.school, assignment=self.assignment)
         self.delegate.assignment = None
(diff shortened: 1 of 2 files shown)


tdowds:
This PR looks good to me.

huxley/api/tests/test_user.py
```diff
@@ -416,3 +416,67 @@ def test_get(self):
...
+
+class DelegateUserDestroyTestCase(tests.DestroyAPITestCase):
+    url_name = 'api:delegate_detail'
+
+    def setUp(self):
+        self.advisor = models.new_user(username='advisor', password='advisor')
+        self.school = models.new_school(user=self.advisor)
+        self.registration = models.new_registration(school=self.school)
+        self.committee = models.new_committee()
+        self.country = models.new_country()
```
kmeht: I don't think you need most of these variables. The test models module usually creates whatever it needs by default, so if you're not gonna be using them in the actual test, no need to manually create.

huxley/api/tests/test_user.py
```diff
@@ -416,3 +416,67 @@ def test_get(self):
...
+
+
+class DelegateUserCreateTestCase(tests.PartialUpdateAPITestCase):
+    url_name = 'api:delegate_detail'
+
+    def setUp(self):
+        self.advisor = models.new_user(username='advisor', password='advisor')
+        self.school = models.new_school(user=self.advisor)
+        self.registration = models.new_registration(school=self.school)
+        self.committee = models.new_committee()
```
kmeht: Same here, can probably get away with not manually creating some of these.

PR #645, opened 2017-12-29
Committee feedback

jaketibbetts:
Added backend support for committee feedback including model, serializes, views, permissions, etc. Provided corresponding unit tests that test critical functions, particularly that delegates can only submit one feedback. 

Definitely could use some feedback on how permissions were implemented.

Done in preparation for adding frontend support for feedback.


huxley/api/permissions.py
```diff
@@ -222,6 +223,40 @@ def has_permission(self, request, view):
...
+
+    def has_permission(self, request, view):
+        user = request.user
+        if user.is_superuser:
+            return True
+
+        method = request.method
+        committee_id = view.kwargs.get('committee', -1)
+        if method == 'POST' and user.is_authenticated() and user.is_delegate(
+        ) and user.delegate.assignment:
```
m-j-mcdonald: Breaking the parentheses here feels awkward; perhaps try splitting at each 'and' instead

huxley/core/migrations/0025_auto_20171105_1635.py
```diff
@@ -0,0 +1,39 @@
...
+# Generated by Django 1.10.7 on 2017-11-05 16:35
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('core', '0024_auto_20170927_2010'), ]
```
m-j-mcdonald: Move this dependency forward past the newer migrations (0026 is the most recent).

huxley/core/admin/committee_feedback.py
```diff
@@ -0,0 +1,14 @@
...
+
+from django.conf.urls import url
+from django.contrib import admin
+from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect
+
+from huxley.core.models import CommitteeFeedback
+
+
+class CommitteeFeedbackAdmin(admin.ModelAdmin):
```
m-j-mcdonald: This file is unnecessary I believe

huxley/api/views/committee_feedback.py
```diff
@@ -0,0 +1,39 @@
...
+class CommitteeFeedbackDetail(generics.CreateAPIView):
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (permissions.CommitteeFeedbackDetailPermission, )
+    serializer_class = CommitteeFeedbackSerializer
+
+    def perform_create(self, serializer):
+        if not self.request.user.is_superuser:
+            delegate = Delegate.objects.get(pk=self.request.user.delegate.id)
+            delegate.committee_feedback_submitted = True
+            #self.request.user.delegate.refresh_from_db()
```
m-j-mcdonald: Remove commented lines like this

huxley/api/views/committee_feedback.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class CommitteeFeedbackDetail(generics.CreateAPIView):
+    authentication_classes = (SessionAuthentication, )
+    permission_classes = (permissions.CommitteeFeedbackDetailPermission, )
+    serializer_class = CommitteeFeedbackSerializer
+
+    def perform_create(self, serializer):
+        if not self.request.user.is_superuser:
+            delegate = Delegate.objects.get(pk=self.request.user.delegate.id)
```
m-j-mcdonald: There shouldn't be a need to query the database here. This should be rewritten as:

`
delegate = self.request.user.delegate
`

huxley/api/tests/test_committee_feedback.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertEqual(response_1.data, {
+            "committee": self.committee.id,
+            "comment": self.params['comment'],
+        })
+
+        #This is how to refresh an object after it was updated from somewhere else
+        self.user.delegate.refresh_from_db()
+        self.assertTrue(self.user.delegate.committee_feedback_submitted)
+        response_2 = self.get_response(self.committee.id, params=self.params)
+        print(response_2)
```
m-j-mcdonald: Don't put print statements into test cases

huxley/api/tests/test_committee_feedback.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertFeedbackPreviouslySubmitted(response_2)
```
m-j-mcdonald: What's going on here?

huxley/api/permissions.py
```diff
@@ -222,6 +223,40 @@ def has_permission(self, request, view):
...
+            if user.delegate.committee_feedback_submitted:
+                raise ValidationError(
+                    {'detail': "Delegate feedback was already submitted."})
```
m-j-mcdonald: I think PermissionDenied would be more appropriate here than a validation error

huxley/core/admin/committee_feedback.py
m-j-mcdonald: The default admins should work just fine; you can omit the second argument to register in the __init__ file

huxley/api/permissions.py
jaketibbetts: Lol yea I agree, I think the autoformatter did that one

huxley/core/admin/committee_feedback.py
jaketibbetts: Done

huxley/api/views/committee_feedback.py
jaketibbetts: Done

huxley/api/views/committee_feedback.py
jaketibbetts: Done

huxley/api/tests/test_committee_feedback.py
jaketibbetts: Done

huxley/api/tests/test_committee_feedback.py
jaketibbetts: This test checks that a delegate can only submit one feedback.

Because we don't store the delegate_id anywhere on the CommitteeFeedback model to maintain anonymous feedback, the only real way to check to see if a delegate has submitted feedback previously is the committee_feedback_submitted field on the Delegate model. This is a check to see that this is being updated properly.

jaketibbetts: We could separate it into separate tests if you think that's necessary? But I don't think it is

huxley/api/permissions.py
jaketibbetts: Shouldn't PermissionDenied be related to permissions issues? This error is raised when a delegate has already submitted feedback and attempts to submit another which doesn't seem like a permissions issue. I also see how it's not really a validation issue either though because it has nothing to do with the data being passed in.

The only reason I put this in the permissions class at all is because I couldn't figure out a way access user.delegate data in the serializer when we could put a validator like in the school serializer. Is there a better way to do this or should be just leave it as PermissionDenied and call it there?

huxley/core/migrations/0025_auto_20171105_1635.py
jaketibbetts: Oops, forgot to commit new dependencies

jaketibbetts: Also what do you mean "move forward"? I don't know a lot about how migrations and dependencies work.

m-j-mcdonald: By move forward I just meant make it dependent on the most recent migration. Its not really proper terminology.

huxley/api/permissions.py
m-j-mcdonald: I think PermissionDenied makes sense here; a delegate isn't allowed to submit more than once. Validation errors are typically for when there's an issue with a piece of the information you're trying to save, not really when you shouldn't be saving at all.

huxley/core/admin/__init__.py
```diff
@@ -9,6 +9,7 @@
 
 from .assignment import AssignmentAdmin
 from .committee import CommitteeAdmin
+from .committee_feedback import CommitteeFeedbackAdmin
```
m-j-mcdonald: Remember to get rid of this line

jaketibbetts: Done

huxley/api/permissions.py
jaketibbetts: Okay sounds good to me

huxley/core/admin/__init__.py
jaketibbetts: That's what made Travis fail this time whoops
screenshot from 2018-01-02 11-49-12

huxley/core/migrations/0025_auto_20171105_1635.py
jaketibbetts: Ok sounds good! Also just for my own knowledge, what's the point of doing this?

jaketibbetts: Looking into it more, guessing it has to do with wanting this migration to appear after the migrations for new features already merged onto the master branch (like the position paper grading you've been working on)

huxley/www/js/entry.js
```diff
@@ -56,6 +57,7 @@ var routes = (
     <Route path="/advisor/roster" component={AdvisorRosterView} />
     <Route path="/chair/attendance" component={ChairAttendanceView} />
     <Route path="/chair/summary" component={ChairSummaryView} />
+    <Route path="/delegate/committee_feedback" component={DelegateCommitteeFeedbackView} />
     <Route path="/delegate/profile" component={DelegateProfileView} />
```
m-j-mcdonald: Last nit I swear; this shouldn't be added until the view actually exists.

huxley/www/js/entry.js
```diff
@@ -20,6 +20,7 @@ var AdvisorProfileView = require('components/AdvisorProfileView');
 var AdvisorRosterView = require('components/AdvisorRosterView');
 var ChairAttendanceView = require('components/ChairAttendanceView');
 var ChairSummaryView = require('components/ChairSummaryView');
+var DelegateCommitteeFeedbackView = require('components/DelegateCommitteeFeedbackView');
```
m-j-mcdonald: Don't reference the view until it exists

huxley/www/js/entry.js
```diff
@@ -56,6 +57,7 @@ var routes = (
     <Route path="/advisor/roster" component={AdvisorRosterView} />
     <Route path="/chair/attendance" component={ChairAttendanceView} />
     <Route path="/chair/summary" component={ChairSummaryView} />
+    <Route path="/delegate/committee_feedback" component={DelegateCommitteeFeedbackView} />
```
m-j-mcdonald: Don't reference the view until it exists

huxley/www/js/components/DelegateView.js
```diff
@@ -21,6 +21,7 @@ var DelegateView = React.createClass({
         <Shaker>
           <div className="navbar rounded-top">
             <NavTab href="/delegate/profile">Profile</NavTab>
+            <NavTab href="/delegate/committee_feedback">Committee Feedback</NavTab>
```
m-j-mcdonald: Wait until the view exists

huxley/www/js/entry.js
jaketibbetts: Oh whoops I didn't mean to commit that :/ darn.

Well I've been hacking at the frontend a lot in the last few days and should be done with that relatively soon anyways so I'll push those soon!

m-j-mcdonald: I can push this as is without these; otherwise you'll have to wait until the frontend exists for the PR to close which will mean dealing with new migrations (there's some code I'm going to push up very soon)

jaketibbetts: Yea works for me

jaketibbetts: Can you do that in github or do you need me to push it up?

m-j-mcdonald: I can do it in Github this time, but generally it's better if you do it on your side

huxley/api/tests/test_committee_feedback.py
```diff
@@ -0,0 +1,123 @@
...
+
+from django.core.exceptions import ValidationError
+
+from huxley.accounts.models import User
+from huxley.api import tests
+from huxley.api.tests import auto
+from huxley.utils.test import models
+import random
+
+
```
m-j-mcdonald: You need to be more comprehensive in this file; you can't just test the parts you want to. Add tests for all users' behavior in all cases (think what would happen if we expect a chair not to be able to create feedback but someone accidentally adds code that does that; this test needs to catch it). Unit tests shouldn't test the current functionality they should test the contract with that part of the code; if the code is supposed to have a certain behavior that behavior needs to be tested even if at no point in the existing code is that behavior used.

huxley/api/tests/__init__.py
```diff
@@ -107,6 +115,11 @@ class CreateAPITestCase(AbstractAPITestCase):
     is_resource = False
 
 
+class PostSingleAPITestCase(AbstractAPITestCase):
```
m-j-mcdonald: Rename this to Create...
Also, why do you need is_resource to be true?

huxley/api/tests/test_committee_feedback.py
jaketibbetts: Yup I'll get to adding these! Shouldn't be too bad because we'll just be returning PermissionDenied on pretty much everything else.

huxley/api/tests/__init__.py
jaketibbetts: In `get_url` method in **tests/__init__.py**

`args = (object_id, ) if self.is_resource else ()`
`return reverse(self.url_name, args=args)`

So `is_resource` must be true or it doesn't generate the url correctly

If you look at **urls.py** the urls for committee_feedback are:

`    url(r'^committee_feedback/?$',`
`        views.committee_feedback.CommitteeFeedbackList.as_view(),`
`        name='committee_feedback_list'),`
`    url(r'^committee_feedback/(?P<committee>[0-9]+)/?$',`
`        views.committee_feedback.CommitteeFeedbackDetail.as_view(),`
`        name='committee_feedback_detail'),`

So you need to have `is_resource` be true so the post request goes to the committee_feedback_detail url.

Now that I'm looking at this again, do you think this needs to be changed?

There's no way to reference a single committee feedback right now. For this feature, that isn't an issue because we only ever retrieve them in a lists for the ChairCommitteeFeedbackView. However if we wanted to add something else that requires being able to reference a single feedback, we don't have that right now.

If we do change it, how do we go to that url to create a feedback without knowing the feedback's ID (Because it hasn't been created in the database)? Could you just pass in a random integer?

Other than that one question, I have no issue changing this and I don't think it would be too much work. Also, it would be more consistent with how the rest of the api works.

jaketibbetts: Okay I messed around with that kind of implementation on my machine and it works completely fine. You just have to pass in a random integer.

Maybe we could set it so that all post requests for single entry use 0 for the pk. All SQL databases start autoincrementing at 1 and all other HTTP requests would access an entry in the db that already exists.

huxley/www/js/lib/ServerAPI.js
```diff
@@ -15,6 +15,10 @@ var ServerAPI = {
     return _post('/api/delegates', {name, email, school});
   },
 
+  createCommitteeFeedback(feedback) {
+    return _post(`/api/committee_feedback/0`, feedback);
```
m-j-mcdonald: There should be a way to post without needing an id value here. It will need a change at the api level though

huxley/www/js/text/DelegateCommitteeFeedbackViewText.md
```diff
@@ -0,0 +1,3 @@
+Thank you for attending BMUN {{ conferenceSession }} {{ firstName }}!  Please leave your feedback for **{{ committee }}**. 
+
+All delegate feedback is anonymous.
```
m-j-mcdonald: Add something here in bold to stress they get to submit only once

huxley/api/permissions.py
```diff
@@ -222,6 +222,47 @@ def has_permission(self, request, view):
...
+       Accept GET request from chair of committee'''
+
+    def has_permission(self, request, view):
+        user = request.user
+        if user.is_superuser:
+            return True
+
+        method = request.method
+        committee_id = request.data.get('committee', -1)
+        feedback_id = view.kwargs.get('pk', None)
```
m-j-mcdonald: It's probably better to have this as -1 instead of None

huxley/api/tests/__init__.py
m-j-mcdonald: I've got a simpler way around this: change the detail url to be a fixed address (say committee_feedback/post). There's actually never a reason to pass pk to the url, since we never use this url to do database lookups (and likely never will).

huxley/www/js/text/DelegateCommitteeFeedbackViewText.md
jaketibbetts: Done

huxley/api/permissions.py
jaketibbetts: Done

huxley/www/js/lib/ServerAPI.js
jaketibbetts: Like we discussed, we'll add a fixed url to handle single post requests. I think it's the cleanest way to do it.

PR #656, opened 2018-02-04
Fixed Registration Error

shaynak:
Changed max values in delegate fees fields and made sure too many delegates can't be registered.


m-j-mcdonald:
Lgtm; I'll merge this in now

PR #674, opened 2018-10-28
Advisor roster

shaynak:
added waiver and position paper checkoffs in advisor roster view


huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -168,10 +170,31 @@ var AdvisorRosterView = React.createClass({
             </Button>
           </td>
         );
+
+        var waiverCheck = '';
```
kmeht: Don't use `var`, use `let` or `const`. Also, this can be a one liner:
```
const waiverCheck = delegate && delegate.waiver_submitted ? '\u2611' : '\u2610';

File: huxley/www/js/components/AdvisorRosterView.js [hunk 2]
```diff
@@ -168,10 +170,31 @@ var AdvisorRosterView = React.createClass({
...
+        if (delegate && delegate.waiver_submitted) {
+          waiverCheck = '\u2611';
+        } else {
+          waiverCheck = '\u2610';
+        }
+
+        var positionPaperCheck = '';
```
kmeht: ditto re: `const`

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -168,10 +170,31 @@ var AdvisorRosterView = React.createClass({
...
+        if (
+          delegate.assignment &&
```
kmeht: If you guys update Babel, you can use the optional property access operator to make this much shorter: https://www.npmjs.com/package/babel-plugin-transform-optional-chaining

huxley/www/js/components/AdvisorRosterView.js
kmeht: ditto re: optional property access

PR #679, opened 2019-01-08
Advisor Roster View

shaynak:
Added position paper + waiver checks to the advisor view.

### DIFF
--- a/huxley/www/js/components/AdvisorRosterView.js
+++ b/huxley/www/js/components/AdvisorRosterView.js
@@ -212,7 +212,6 @@ var AdvisorRosterView = React.createClass({
           </td>
         );
 
-
         const waiverCheck =
           delegate && delegate.waiver_submitted ? '\u2611' : '\u2610';
 
@@ -222,9 +221,6 @@ var AdvisorRosterView = React.createClass({
           assignments[delegate.assignment - 1].paper.file
             ? '\u2611'
             : '\u2610';
-
-
-
 
         return (
           <tr>
--- a/huxley/www/js/components/AdvisorRosterView.js
+++ b/huxley/www/js/components/AdvisorRosterView.js
@@ -179,6 +179,12 @@ var AdvisorRosterView = React.createClass({
     var assignments = this.state.assignments;
     var disableEdit = _checkDate();
 
+    var assignment_ids = {}
+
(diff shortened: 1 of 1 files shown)


huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -204,10 +211,23 @@ var AdvisorRosterView = React.createClass({
             </Button>
           </td>
         );
+
+        const waiverCheck =
+          delegate && delegate.waiver_submitted ? '\u2611' : '\u2610';
+
+        const positionPaperCheck =
+          delegate.assignment &&
```
jaketibbetts: Put an extra null check here for assignments[delegate.assignment - 1]

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -204,10 +211,23 @@ var AdvisorRosterView = React.createClass({
...
+          assignments[delegate.assignment - 1].paper &&
```
jaketibbetts: The use of assignments[delegate.assignment - 1] only works as intended if
id of assignment object from server == key of that assignment object in the "assignments" variable

For example if you got an assignment object from the server with id = 4, your assignments variable would look like
assignments[0] = assignment(id=4)
and your call
assignments[delegate.assignemnt - 1] == assignments[3]
which would be undefined

Try adding dummy assignments in your test env to increment to PK in the DB a little bit and then test with those assignments

huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -43,6 +44,9 @@ var AdvisorRosterView = React.createClass({
     return {
       delegates: DelegateStore.getSchoolDelegates(schoolID),
       registration: RegistrationStore.getRegistration(schoolID, conferenceID),
+      assignments: AssignmentStore.getSchoolAssignments(schoolID).filter(
```
jaketibbetts: This looks great!

jaketibbetts:
Can you also add these changes to the AdvisorProfileView as well? It needs to be updated there

jaketibbetts:
Closing this because I added my own fix in #680 and pushing that to master. All the old commits from this are making it in though.

PR #683, opened 2019-02-04
Advisor papers

shaynak:
This is a very large PR

changes made:
- Added a tab for advisors to see their students' position paper scores + for them to download both original and graded papers.
- Added a field to the PositionPaper model to store graded papers + made changes necessary for that in other parts of code (updated delegate view so they can download graded paper once it's available; made sure chairs were submitting to the graded_file category; etc.)


jaketibbetts:
Just btw for the graded paper paper pop up box. That's added there on purpose because the chair needs to refresh the page to see the updates to the paper.

jaketibbetts:
Updating this with my own changes in #685. Thanks for all the amazing work @shaynak

Issue #713, opened 2020-02-24
Update the Conference model to include more information


shaynak:
Position paper deadlines, delegate edit deadlines, etc. They shouldn't be something Tech has to edit every year -- should be part of the Conference model. Also any dates in AdvisorChecklistDelegateFeeText.md.

The Conference model should also be updated to represent the current Phase 1/2/3 registration system.

This also includes https://github.com/bmun/huxley/issues/710.

srisainachuri:
Still contains hardcoded dates. Hardcoded emails should also be removed and added to the conference model.

PR #718, opened 2020-08-18
Stage - Updating Conference

mathildepm:
#713 
Adding more fields to conference and having those fields be reflected on the front end


huxley/www/js/components/AdvisorProfileView.js
```diff
@@ -23,7 +23,7 @@ const Table = require('components/core/Table');
 const TextInput = require('components/core/TextInput');
 const TextTemplate = require('components/core/TextTemplate');
 const User = require('utils/User');
-const _handleChange = require('utils/_handleChange');
+const _handleChange = require('utils/_handleChange')
```
shaynak, line 26: why is the semicolon gone

shaynak (approved):
lgtm, just fix that missing semicolon!

PR #719, opened 2020-11-26
Fix for failed assignment uploads

mathildepm:
Now when assignment upload fails, all position paper objects that were created in the process are deleted.


shaynak (approved):
lgtm!

PR #724, opened 2021-01-07
Migration updates

mathildepm:
Changed all emails sent from tech to no-reply, changed http html to https and updated delegate account creation email


shaynak (approved):
lgtm

Issue #712, opened 2020-02-19
Google sheets integration for admin panel


shaynak:
Would make External/Admin's lives much easier if there was a Google Sheets integration that allowed us to write data to Sheets. It shouldn't be too difficult to implement, since the Sheets API is nice, but I haven't looked too far into it. Process would probably involve modifying the `huxley/core/api` + relevant models to append to spreadsheet as updates come in.

PR #730, opened 2021-02-02
Google Sheets Integration

shaynak:
#712: Adds a "Sheets" button to the admin view that will allow syncing to a Google sheet.

Steps that need to be done locally / on the server to make sure this works:

1. Make sure all packages are installed!
2. Create a Google spreadsheet with the following sheets: `Assignments`, `Feedback`, `Delegates`, `Registration`, `Schools`
2. Add a variable called `SHEET_ID` to `settings/local.py` with the spreadsheet id (alphanumeric string in the URL to the sheet)
3. Create a service account by [following the instructions here](https://developers.google.com/identity/protocols/oauth2/service-account#creatinganaccount) and place the service account JSON file at `huxley/service.json`. Create a new project for Huxley when doing so; no need to complete the optional steps.
4. Share the spreadsheet created in step (2) with the service account -- make sure to give it edit access!

Clicking the resulting "Sheets" buttons will automatically update the spreadsheet for that model, allowing an easier transfer of data than downloading CSVs.

This also updates two sub-dependencies (because Google Auth relies on a more recent version of them) but it shouldn't be an issue.


mathildepm (approved):
lgtm!

PR #742, opened 2021-02-27
Conference hotfix

mathildepm:
More control from the backend over polling intervals and refresh thresholds. 
Also fixed some logic for polling that was resulting in potentially unending GETs
A small addition - removing the chair's ability to see numerical ratings.


shaynak (approved):
LGTM, hopefully this fixes our issues

Issue #734, opened 2021-02-18
Ignore Headers on Secretariat CSV Upload


mathildepm:
Currently, uploading a csv will error if the csv has headers. Change this!

PR #762, opened 2021-06-30
Fixed starter issue: CSV file headers.

kotrynav:
#734 This issue prevented errors when headers in CSV files were present. Also added HTML error messages on failed uploads to specify which rows could not be matched to existing schools or committees.

### DIFF
--- a/huxley/core/admin/committee.py
+++ b/huxley/core/admin/committee.py
@@ -4,7 +4,7 @@
 import csv
 
 from django.conf.urls import url
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.urls import reverse
 from django.http import HttpResponseRedirect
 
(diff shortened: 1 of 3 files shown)


huxley/core/admin/committee.py
```diff
@@ -17,14 +17,23 @@ def load(self, request):
...
+                    com = Committee(name=row[0],
+                                    full_name=row[1],
+                                    delegation_size=int(row[2]),
+                                    special=special)
+                    com.save()
+                else:
+                    failed_uploads.append(row)
+        if failed_uploads:
+            messages.error(
+                request, 'Not all secretariat members could upload. These rows failed: '
```
mathildepm, line 35:
```suggestion
                request, 'Not all Committees could upload. These rows failed: '
```

mathildepm: JK let's just revert this file

mathildepm (commented):
LGTM! 
Idk why some tests are running so we'll see when you push again. 
Let's revert committee.py to what it was before since we aren't worried about that:
https://dev.to/lofiandcode/git-and-github-how-to-revert-a-single-file-dha

mathildepm:
ALSO; let's update the PR description so it describes more accurately what the PR is doing 
- Reduces errors due to CSV headers 
- Prevents entire failures of uploads if a single row fails - instead tracks failed rows.

srisainachuri:
Here's a reference to the commit in huxley notes app in delegate.py for the email check:
https://github.com/bmun/huxley/commit/9d9859fa0ddef441e1bb05881e34cc2c60dab3bf

PR #765, opened 2021-07-21
Unique country

erwang01:
Enforces country uniqueness in django db model.

TODOs left:
- [x] update unit tests, currently many non unique failures are occurring.
- [x] add non unique unit test


srisainachuri (approved):
Looks good to me!

Issue #764, opened 2021-07-20
Fix file upload bug in DelegatePaperView


srisainachuri:
For some files, especially larger ones, uploading position papers will error and not successfully upload. Most likely cause of issue is history.redirect refreshes the page before the upload actually finishes.

erwang01:
We can consider moving [history.redirect](https://github.com/bmun/huxley/blob/da7a18934cc302db447821a8447e6f62240aa9e0/huxley/www/js/components/DelegatePaperView.js#L216) to inside the upload callback function.
https://github.com/bmun/huxley/blob/da7a18934cc302db447821a8447e6f62240aa9e0/huxley/www/js/components/DelegatePaperView.js#L220-L222

This way we only redirect upon success. However #785 is probably a better way of doing things.

srisainachuri:
Moving the history.redirect works! #785 is probably cleaner, currently looking into why the [existing listener](https://github.com/bmun/huxley/blob/da7a18934cc302db447821a8447e6f62240aa9e0/huxley/www/js/components/DelegatePaperView.js#L59) doesn't work as intended.

erwang01:
_handleSuccess doesn't seem to change the state at all. Try moving this into _handleSuccess? https://github.com/bmun/huxley/blob/da7a18934cc302db447821a8447e6f62240aa9e0/huxley/www/js/components/DelegatePaperView.js#L212-L214
I'm not entirely sure what the purpose of resetting this state is but I guess it lets the user upload a new file if they want to change it?

Also what does a listener do again?

Issue #761, opened 2021-06-26
In DelegatePaperView, show title of uploaded paper


mathildepm:
When delegates upload papers to Huxley. They can click "Download Paper" to confirm they uploaded the right file but they still have to go through the process of downloading a file they already have. IT would be nice to display the filename next to the button so they can be sure it was the right file/it has been uploaded.

PR #766, opened 2021-09-03
Add filename to paper view

erwang01:
Screen Shot 2021-09-02 at 9 55 45 PM
Adds the filename such that users can see which file was uploaded. 

WARNING: Unfortunately the file name currently displayed is NOT the same file name as the one the user uploaded. Instead its the file name in our system.

Test user uploaded `chap3.pdf`, it is saved and displayed as `chap3_oUZqjc7.pdf`


mathildepm (approved):
#761 
LGTM! 
I tested it out a bit and seems like there are still some issues with Huxley saying "something went wrong" when we upload a larger paper, even though it works! #764 aims to address this - but it wasn't a problem last year :( - BUT this is irrelevant to this functionality so should be good to merge.

PR #775, opened 2021-10-23
Added copy email button and fixed table alignment in chair delegate page

skberkeley:


huxley/www/js/components/ChairDelegateEmailView.js
```diff
@@ -90,29 +92,35 @@ class ChairDelegateEmailView extends React.Component {
...
-          </div>
+          <Table
+            emptyMessage="You don't have any delegates in your committee."
+            isEmpty={!this.state.delegates.length}
+          >
+            <thead>
+              <tr>
+                <th>Assignment</th>
+                <th>Email
+                  <Button
```
srisainachuri, line 105: Functionality seems great! This is more stylistic, but can you move the button outside of the table? You should still be able to access the data from the store as you did, but this will remove the empty column that appears since the button is currently a title of a column.

srisainachuri (changes requested):
Looks good overall! Just added 1 comment where it would be nicer if the button was outside of the table, so there isn't a giant empty column. This might also mean you don't need the table spacing width of 50% you added.

srisainachuri (approved):
For purposes of PR, styling is fine though so we can merge this.

srisainachuri (approved):

Issue #599, opened 2017-07-15
Split requirements.txt into two


kmeht:
**NOTE: This is a temporary solution until [pipfile](https://github.com/pypa/pipfile) is stable, but who knows how long that could take.**

One of the issues that bothers me with pip's dependency management is that transitive dependencies (i.e., dependencies of dependencies) are indistinguishable from the project's top-level dependencies. For example, our `requirements.txt` is littered with dependencies we clearly didn't add, and it's harder to see what the project actually requires.

This proposes we split this into two files: `requirements.txt` and `requirements.lock.txt`. The former will be edited by hand, and only contain the top-level dependencies of the project (e.g. Django, Django Rest Framework, Fabric, etc). The latter will be automatically generated via the command `pip freeze > requirements.lock.txt`, and contain every dependency, transitive included.

Here's a sample workflow:
- When a developer first checks out the project, they install the dependencies from the lock file, via `pip install -r requirements.lock.txt`.
- When a developer wants to upgrade the dependencies of the project (e.g. upgrading to Django 1.10), they edit `requirements.txt` by hand and then run `pip install -r requirements.txt`.
- When everything is tested, they update the lock file via `pip freeze > requirements.lock.txt`.

PR #778, opened 2021-12-01
updating requirements

srisainachuri:
- removed unnecessary dependencies and updating versions using pip freeze
- should resolve pip install errors


kmeht:
This might be a good time to address #599.

srisainachuri:
> This might be a good time to address #599.

good point will look into this

mathildepm (approved):
LGTM! Approving for now as a temporary way to make sure installation is smoother. Although Kunal is right that we should look into implementing #599 for future ease of use.

PR #791, opened 2022-01-17
Password reset

### LINKED ISSUE #758: Improve Password Reset flow
Two things 
1. The password reset email is incredibly (almost hilariously) problematic. It only says 

> The password for user Jeremy_8392 has been reset to QsZktha5dV.
> Thank you for using Huxley!

And the period at the end of the password is there for grammatical purposes, but ends up being included when delegates try the new password. 
Step 1 is to update the email to be more clear. Take a look at the email we send when account are first created. It might be more clear to have a set up like "something something Berkeley Model UN" and then "Username: testuser // Password: cheese" etc

2. Advisors will often click "reset password" without really thinking. This can confuse delegates if they already have their password and it gets reset without them knowing. It can also just been ineffective if the advisor has put the wrong email. Before enacting the password reset, have clicking the button make a pop up appear with details one what is about to happen. "This will send an email to <delegateemail> with new login information. Are you sure this is the correct email? Note that school organizational emails sometimes have emails from outside parties blocked" or something like that. This should allow them to either proceed or go back.

erwang01:
Resolves #758 

Email content changed but Untested.

Advisor confirmation is as follows:
Confirmation Screen
Screen Shot 2022-01-16 at 5 02 16 PM
followed by an alert if the password was successfully changed.
Screen Shot 2022-01-16 at 4 59 42 PM

### DIFF
--- a/huxley/www/js/components/AdvisorRosterView.js
+++ b/huxley/www/js/components/AdvisorRosterView.js
@@ -259,10 +259,7 @@ class AdvisorRosterView extends React.Component {
                   size="small"
                   onClick={e => 
                     window.confirm(`Are you sure you wish to reset the password for ${delegate.name}? The delegate will no longer be able to log in with the old password. Please ensure that the delegate's email is entered correctly: ${delegate.email}`) 
-                    && this._handleDelegatePasswordChange.bind(
-                      this,
-                      delegate
-                  )()}
+                    && this._handleDelegatePasswordChange(delegate)}
                 >
                   Reset Password
                 </Button>
--- a/huxley/www/js/components/AdvisorRosterView.js
+++ b/huxley/www/js/components/AdvisorRosterView.js
@@ -257,10 +257,12 @@ class AdvisorRosterView extends React.Component {
                 <Button
                   color="yellow"
                   size="small"
-                  onClick={this._handleDelegatePasswordChange.bind(
-                    this,
-                    delegate
-                  )}
+                  onClick={e => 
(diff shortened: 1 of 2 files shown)


huxley/www/js/components/AdvisorRosterView.js
```diff
@@ -257,10 +257,12 @@ class AdvisorRosterView extends React.Component {
                 <Button
                   color="yellow"
                   size="small"
-                  onClick={this._handleDelegatePasswordChange.bind(
-                    this,
-                    delegate
-                  )}
+                  onClick={e => 
+                    window.confirm(`Are you sure you wish to reset the password for ${delegate.name}? The delegate will no longer be able to log in with the old password. Please ensure that the delegate's email is entered correctly: ${delegate.email}`) 
+                    && this._handleDelegatePasswordChange.bind(
```
kmeht, line 262: You don't need to bind anymore, that was to create a new function with the arguments already bound. Now that we're in a lambda, just call the function normally.

erwang01: Ah I see, thanks! Should we replace all binds with lambdas in general? I guess this code was written prior to lambdas being a thing?

Seems this stack post recommends lambdas over binds but maybe theres a reason to leave them be?
https://stackoverflow.com/questions/42117911/lambda-functions-vs-bind-memory-and-performance

srisainachuri (approved):
LGTM!

PR #794, opened 2022-02-02
Waivers api

srisainachuri:
Initial PR for waivers API integration. This feature permits integration with the Smartwaiver waiver management console using Celery periodic tasks to automatically retrieve waivers at set intervals and update Delegate objects with waiver status. Celery integration also allows for periodic tasks, which can be expanded to other functionality such as Google sheets sync. 

Server Integration Guide:

1. Make sure necessary dependencies are installed from `requirements.txt`. This includes installing the python packages `celery`, `django-celery-results`, and `django-celery-beat`.
2. Install and run the [rabbitmq](https://www.rabbitmq.com/install-debian.html) message-broker service
3. Configure, [daemonize](https://docs.celeryproject.org/en/stable/userguide/daemonizing.html#usage-systemd), and start a celery service with a single worker (to avoid concurrent waiver API calls)
4. Configure, [daemonize](https://docs.celeryproject.org/en/stable/userguide/daemonizing.html#usage-systemd), and start django celery beat. Make sure to specify the flag ` --scheduler django_celery_beat.schedulers:DatabaseScheduler` to connect celery beat to the django backend.
5. Add a variable called `SMARTWAIVER_API_KEY` to `settings/local.py` and retrieve an API key from the Smartwaiver console (instructions [here](https://api.smartwaiver.com/docs/v4/)).  

Once all server steps are completed, periodic tasks can be added and configured in the Admin view, with the available periodic tasks defined in `huxley/core/tasks.py`. With the task `poll_waiver`, waivers that successfully match delegates will update the corresponding `Delegate` object and waivers that fail to match delegate will generate logs of the model `WaiverLog`.

Future TODOs:
- [ ] Better handle process_waiver response under Delegate in core/models.py
- [ ] Unit Tests


mathildepm (approved):
Looks great! 
In the PR description, can you add a list of steps for how to modify local files/envs to work with this integration? Similar to what Shayna did in #730

srisainachuri:
> Looks great! In the PR description, can you add a list of steps for how to modify local files/envs to work with this integration? Similar to what Shayna did in #730

Updated!

PR #802, opened 2022-02-24
Hotfixes

srisainachuri:
* Graded papers were using the uploaded paper filename and extension. This was causing problems especially if the delegate uploaded paper was a different format such as .docx than the graded paper which may have been .pdf. 

* Rejected assignments should not be factored in for the Advisor checklist when checking that all position papers are submitted

### DIFF
--- a/huxley/www/js/components/PaperGradeTable.js
+++ b/huxley/www/js/components/PaperGradeTable.js
@@ -38,7 +38,9 @@ class PaperGradeTable extends React.Component {
           : null;
       var fileNames = paper.file.split("/");
       var fileName = fileNames[fileNames.length - 1];
-      var gradedName = gradedHrefData ? "graded_" + fileName : null;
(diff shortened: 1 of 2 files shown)


huxley/www/js/components/PaperGradeTable.js
```diff
@@ -38,7 +38,9 @@ class PaperGradeTable extends React.Component {
           : null;
       var fileNames = paper.file.split("/");
       var fileName = fileNames[fileNames.length - 1];
-      var gradedName = gradedHrefData ? "graded_" + fileName : null;
+      var gradedFileNames = paper.graded_file.split("/");
+      var gradedFileName = gradedFileNames[gradedFileNames.length - 1];
+      var gradedName = gradedHrefData ? "graded_" + gradedFileName : null;
```
erwang01, line 43: why is there a graded_ prepending to gradedFileName?

PR #803, opened 2022-02-25
Hotfixes

srisainachuri:
fixing graded file link issues to delegate and advisor views + fixing null pointers for resubmitted graded papers

### DIFF
--- a/huxley/www/js/components/AdvisorProfileView.js
+++ b/huxley/www/js/components/AdvisorProfileView.js
@@ -152,8 +152,9 @@ class AdvisorProfileView extends React.Component {
       for (var i = 0; i < assignments.length; i++) {
         if (
           assignments[i] == null ||
-          assignments[i].paper == null ||
-          assignments[i].paper.file == null
+          (assignments[i].rejected == false &&
+          (assignments[i].paper == null ||
+          assignments[i].paper.file == null))
         ) {
           positionPapersTurnedIn = "\u2610";
           break;
--- a/huxley/www/js/components/AdvisorProfileView.js
+++ b/huxley/www/js/components/AdvisorProfileView.js
@@ -133,7 +133,7 @@ class AdvisorProfileView extends React.Component {
     if (
       registration &&
       registration.registration_fee_paid &&
-      registration.delegate_fees_paid == registration.delegate_fees_owed
+      registration.delegate_fees_paid >= registration.delegate_fees_owed
     ) {
       allFeesPaid = "\u2611";
     } else {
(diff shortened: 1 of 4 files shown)


huxley/www/js/components/AdvisorProfileView.js
```diff
@@ -133,7 +133,7 @@ class AdvisorProfileView extends React.Component {
     if (
       registration &&
       registration.registration_fee_paid &&
-      registration.delegate_fees_paid == registration.delegate_fees_owed
+      registration.delegate_fees_paid >= registration.delegate_fees_owed
```
erwang01, line 136: Do we provide refunds or what not? This is the advisor checklist so probably not the right view but curious if somewhere we have a flag for a yes we are refunding you in the future.

huxley/www/js/components/PaperSubmissionTable.js
```diff
@@ -24,11 +24,17 @@ class PaperSubmissionTable extends React.Component {
...
+      var gradedHrefData =  
+          paper.graded &&
+          paper.graded_file &&
+          graded_files[paper.id]
+            ? url.createObjectURL(graded_files[paper.id])
+            : null;
       var fileNames = paper.file.split("/");
       var fileName = fileNames[fileNames.length - 1];
+      var gradedFileNames = gradedHrefData ? paper.graded_file.split("/") : null;
+      var gradedFileName = gradedFileNames ? gradedFileNames[gradedFileNames.length - 1] : null;
```
erwang01, line 36: Over here, what if gradedFileNames is empty list, so `gradedFileName` is null then the next line we try to add "graded_" to null?

I don't think it will happen unless `paper.graded_file` is empty string or something similarly odd but the code just looks a bit funky.

erwang01 (commented):
LGTM. Not tested on stage.

Lots of reused code segments makes me think we can shove some parts of this into a react component on its own and pass the `paper` object and `graded_files[paper.id]` into the component.

Also I wonder what JS code convention is between the repeated gradedHrefData ? and a giant if {} block.

mathildepm (approved):
LGTM! Should consider Eric's suggested for later bug fixes since Paper grading is a nightmare.

huxley/www/js/components/AdvisorProfileView.js
srisainachuri: We do provide refunds but until they are processed, it's for the treasurer's records that the fees paid might be greater than fees owed.

huxley/www/js/components/PaperSubmissionTable.js
srisainachuri: gradeFileName shouldn't be null if gradedHrefData isn't null so in this case it should be ok

srisainachuri:
> LGTM. Not tested on stage.
> 
> Lots of reused code segments makes me think we can shove some parts of this into a react component on its own and pass the `paper` object and `graded_files[paper.id]` into the component.
> 
> Also I wonder what JS code convention is between the repeated gradedHrefData ? and a giant if {} block.

This is a good point. A lot of the reused code could be condense into a separate component since it probably reflects being written at different times.

PR #810, opened 2022-06-21
Issue 797: Created a seperate markdown for before the waiver is available so link is not shown

sonikavuyyuru:
Added in logic to check if the waiver is already available based on the waiver's availability date. Displays the appropriate text, only showing the waiver link if it is past the date that the waiver is available.


srisainachuri (changes requested):
Some minor things to look at, but overall logic looks good! Let's also look into why the integration tests are failing before merging

PR #812, opened 2022-09-18
Sonika payment type

sonikavuyyuru:
Updated frontend to 71, payment type question, and payment type unittest changes


srisainachuri (approved):
LGTM!

PR #813, opened 2022-09-19
Invoice automation

sonikavuyyuru:
Sora's invoice automation for registration


huxley/invoice_automation/src/authentication/authenticator.py
```diff
@@ -0,0 +1,67 @@
...
+from intuitlib.client import AuthClient
+from intuitlib.enums import Scopes
+
+
+class Authenticator:
+    # valid options: sandbox, production
+    ENVIRONMENT = "production"
+    REDIRECT_URI = "http://localhost:8000"
+    HOST_NAME = "localhost"
+    SOCKET_NUMBER = 8000
```
srisainachuri, line 15: Does this socket number setting work regardless of what server we are deploying to? Wondering if this is a local test-specific decision or if this works fine for production as well.

huxley/invoice_automation/src/authentication/authenticator.py
```diff
@@ -0,0 +1,67 @@
...
+        server_address = (self.HOST_NAME, self.SOCKET_NUMBER)
+        server = HTTPServer(server_address=server_address, RequestHandlerClass=self.CallbackHandler)
+        server.auth_client = None
+        server.client_id = self.client_id
+        server.client_secret = self.client_secret
+        server_daemon = threading.Thread(target=server.serve_forever)
+        server_daemon.daemon = True
+        server_daemon.start()
+        print(auth_url)
+        while True:
```
srisainachuri, line 58: Is this a continuously running background process as soon as Huxley is started? When does the thread that is spawned terminate and when does this loop terminate in the context of the main app?

huxley/invoice_automation/src/handler/fc_handler.py
```diff
@@ -0,0 +1,112 @@
...
+
+def main():
+    """Shows basic usage of the Sheets API.
+        Prints values from a sample spreadsheet.
+        """
+    creds = None
+    # The file token.json stores the user's access and refresh tokens, and is
+    # created automatically when the authorization flow completes for the first
+    # time.
+    token_path = '../../google_auth_resources/token.json'
```
srisainachuri, line 36: can you move this to django settings similar to how the sheets integration already does it?

huxley/invoice_automation/src/handler/fc_handler.py
```diff
@@ -0,0 +1,112 @@
...
+from huxley.invoice_automation.src import handler
+from huxley.invoice_automation.src.model.address import Address
+from huxley.invoice_automation.src.model.conference import Conference
+from huxley.invoice_automation.src.model.registration import Registration
+from huxley.invoice_automation.src.model.school import School
+
+SCOPES = ['https://www.googleapis.com/auth/spreadsheets.readonly']
+
+# The ID and range of a sample spreadsheet.
+SAMPLE_SPREADSHEET_ID = '1w-Xt-LmWnWvmzDJ6EAp-_I6WCJe9XDlHyAcWOsZ0ZWI'
```
srisainachuri, line 22: what is the purpose of this sample sheet we're writing to here? if the sheet is an actual sheet, the sheet id should also be moved to django settings

huxley/invoice_automation/src/model/conference.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum
+
+
+class Conference(Enum):
+    """
+    Enum representing different conferences
+    Used in Registration, and to determine which line items to use when creating invoices
+    """
+    BMUN71 = "BMUN 71"
```
srisainachuri, line 9: can you instead parse the current conference from the conference model associated with registrations? additionally, we don't support registration through FC so that isn't need. we want to avoid extra hard coding of information that can be parsed from models

huxley/invoice_automation/src/util/quick_books_utils.py
```diff
@@ -0,0 +1,276 @@
...
+
+    detail = SalesItemLineDetail()
+    detail.ItemRef = cc_fee_item.to_ref()
+    detail.ServiceDate = datetime.date.today().isoformat()
+    line.SalesItemLineDetail = detail
+
+    return line
+
+
+END_DATES = {
```
srisainachuri, line 208: This file seems to have a lot of hardcoding of dates and is specific to BMUN 71. Can we generalize this again based on the conference model?

huxley/settings/main.py
```diff
@@ -145,3 +145,10 @@
 
 # Celery Configuration Options
 CELERY_RESULT_BACKEND = 'django-db'
+
```
srisainachuri, line 148: remove these keys and set them to null for the code base, we will set private keys in prod

huxley/core/constants.py
```diff
@@ -35,6 +35,7 @@ class ProgramTypes(Constants):
     CLUB = 1
     CLASS = 2
 
+
```
srisainachuri, line 38: extra space can remove

srisainachuri (changes requested):
Left some comments on individual files.
High Level:

- Lots of files seem to be hard coded for BMUN 71. These need to be generalized so they are easy to update each year and so future changes on the module are easier.
- This module is designed to start up when Huxley starts. Can you add some kind of toggle that would allow us to quickly turn off the automation if needed? This can be something like a checkbox in the conference model.

huxley/api/views/register.py
```diff
@@ -50,6 +59,59 @@ def register(self, request, *args, **kwargs):
...
+                         payment_method: PaymentMethod):
+    school = invoiceSchool(school_name, email, phone_numbers, address)
+    registration = invoiceRegistration(
+        school=school,
+        num_delegates=num_delegates,
+        conference=invoiceConference.BMUN71,
+        registration_date=datetime.date.today(),
+        payment_method=payment_method
+    )
+    handler.handle_registration(registration)
```
srisainachuri, line 117: Is there any error handling here if the module fails to make the API call to QuickBooks successfully? Registration should not fail if the API call does not go through and we should allow for requiring a manual invoice sent. On this note, it might be worth noting on registrations directly if the invoice was already sent or not so we can easily track which registrations need manual invoices.

huxley/invoice_automation/src/authentication/authenticator.py
skberkeley: This module was implemented when I thought we were going to do the OAuth2 flow on the huxley server rather than copying the tokens during deployment. However, the socket number was chosen when I was doing local testing, but I don't expect this module to be used unless we change the current deployment flow.

huxley/invoice_automation/src/authentication/authenticator.py
skberkeley: This code is currently never run by Huxley. It should not be run as part of Huxley until further changes are made since it's designated as a daemon, so it only terminates when the main app process terminates.

huxley/invoice_automation/src/handler/fc_handler.py
skberkeley: Honestly I'll remove fc_handler from the commit, since it's not actually relevant to Huxley.

huxley/invoice_automation/src/model/conference.py
skberkeley: FC was needed for fc_handler, since inside QuickBooks itself we have a separate item. Will make BMUN change though

huxley/invoice_automation/src/util/quick_books_utils.py
skberkeley: Hardcoding based on the conference model means that these won't be available when trying to run scripts with the invoice automation library offline. Can we make this change in a future change once all registration invoices are fully automated?

huxley/core/constants.py
skberkeley: See https://peps.python.org/pep-0008/#blank-lines

huxley/api/views/register.py
skberkeley: Sure, I'll add some error handling

skberkeley: I actually have some doubts with regard to adding error handling at the moment, since we don't have proper logging implemented. Since the invoice handler is called after the registrations are serialized, as far as I'm aware, the unhandled exception won't impact functionality other than not returning a 200 response. However, it has the benefit that Django will log the exception along with a stack trace, making it easier to debug beyond knowing that invoicing failed for a particular registration.

huxley/settings/main.py
skberkeley: Done

huxley/invoice_automation/src/authentication/authenticator.py
srisainachuri: Okay that sounds good! Can we remove this part for now then?

huxley/invoice_automation/src/authentication/authenticator.py
srisainachuri: If it's not run do we need to keep this in the module still or can we remove it since we aren't running as a daemon?

huxley/invoice_automation/src/handler/fc_handler.py
srisainachuri: sg

huxley/settings/main.py
erwang01: Not that its really going to happen but if the keys were already in the git history, you may want to regenerate the keys on Inuit's side.

huxley/core/constants.py
erwang01: We may want to have a separate check for linting. Either at the PR stage or at the push stage.

Lots of the linting errors can be autofixed by off the shelf software.

huxley/api/views/register.py
```diff
@@ -50,6 +59,59 @@ def register(self, request, *args, **kwargs):
             registration_serializer.is_valid(raise_exception=True)
             registration_serializer.save()
 
+        if Conference.get_current().invoicing_enabled and handler is not None:
```
erwang01, line 62: Why will `handler` ever be `None`?

huxley/api/views/register.py
```diff
@@ -50,6 +59,59 @@ def register(self, request, *args, **kwargs):
...
+            school_data = user_data['school']
+            address = Address(
+                line1=school_data['address'],
+                line2='',
```
erwang01, line 66: is school's address guaranteed to be in only 1 line?

like for example if they have a particular room number for the teacher's classroom, how is that handled?

erwang01 (commented):
partial review. Concern on line 66 of register.py is probably important to consider.

huxley/invoice_automation/src/authentication/authenticator.py
skberkeley: Yeah I'll remove the authentication module from the commit too

huxley/invoice_automation/src/authentication/authenticator.py
skberkeley: Resolving bc I'm removing the authentication module entirely

huxley/invoice_automation/src/model/conference.py
skberkeley: Resolving: will remove FC once I put stuff in a separate repo

huxley/invoice_automation/src/util/quick_books_utils.py
skberkeley: Resolving: Changes will be made in separate PR

huxley/core/constants.py
skberkeley: @srisainachuri

huxley/api/views/register.py
skberkeley: If authentication fails, the exception is handled by setting `handler` to `None`. This also allows the existing unit test suite to not fail, since we won't try to make QuickBooks API calls with an unauthenticated handler

huxley/api/views/register.py
skberkeley: I originally implemented Address with two address lines, but it wasn't until I integrated with `register.py` that I realized that Huxley itself only gives one line for street addresses

huxley/api/views/register.py
skberkeley: @srisainachuri thoughts?

huxley/core/constants.py
srisainachuri: alright double space matches the style guide so we can keep

huxley/api/views/register.py
srisainachuri: You're right in that the registration gets serialized properly, but returning a non-200 response will make it seem like it errored on the advisor end so their screen will suggest they didn't response properly. Can you try catch it to ensure that the response is still 200 if the registration was serialized properly? Additionally, for invoices tracking purposes, it might be worthfile to add a checkbox to the registration model if invoices were sent or not, so you can easily manually send the ones that need to be sent instead of parsing through django stack traces

srisainachuri: Here's the [registration handler](https://github.com/bmun/huxley/blob/cbd2c2284ae1c1e7275c1855586647fd12f70951/huxley/www/js/components/RegistrationView.js#L545) in the frontend:

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+from quickbooks.objects.customer import Customer
+
+from huxley.invoice_automation.src.model.conference import Conference
+from huxley.invoice_automation.src.model.fee_type import FeeType
+from huxley.invoice_automation.src.model.payment_method import PaymentMethod
+from huxley.invoice_automation.src.model.registration import Registration
+from huxley.invoice_automation.src.model.school import School
+from huxley.invoice_automation.src.util import quick_books_utils
+
+# TODO: Replace with prod versions during deployment
```
srisainachuri, line 18: Is this todo done since it seems ENVIRONMENT is set to production?

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+# Should probably move these to settings/main.py at some point
```
srisainachuri, line 19: Leaving an unresolved thread here so we can do this in the future.

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def query_schools_as_customers(self, school_names: List[str]) -> List[School]:
+        """
+        Looks for existing customers with the same display names as the passed school names
+
+        :param school_names: List of strings containing school names
+        :return: List of school objects corresponding to customers which were found
+        :raises QuickbooksException:
+        """
+        # qbCustomers = Customer.choose(school_names, DISPLAY_NAME, self.quickbooks_client)
```
srisainachuri, line 117: can this commented out be removed? similar for other commented out code noted below in this fine

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+        :return:
+        """
+        existing_customer = self.get_customer_from_school(school)
+        new_customer = quick_books_utils.get_customer_from_school(school)
+
+        if existing_customer:
+            new_customer.Id = existing_customer.Id
+            new_customer.SyncToken = existing_customer.SyncToken
+
+        # new_customer.save(qb=self.quickbooks_client)
```
srisainachuri, line 137: remove commented out code

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+        return new_customer
+
+    def get_customer_from_school(self, school: School) -> Customer:
+        """
+        Gets the customer object corresponding to the passed school by matching display name
+
+        :param school:
+        :return:
+        """
+        # customer_matches = Customer.choose([school.school_name], field=DISPLAY_NAME, qb=self.quickbooks_client)
```
srisainachuri, line 149: commented out code

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        Queries Quickbooks for invoices which were billed to the passed customer
+        Filters queries for those from the past year
+        :param customer_ref:
+        :return: non-empty list of Invoices or None
+        """
+        # construct query
+        query = construct_invoice_query(customer_ref)
+        # issue query
+        # invoices = Invoice.query(query, qb=self.quickbooks_client)
```
srisainachuri, line 189: commented out code

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+    def query_line_items_from_conference(self, conference: Conference) -> List[Item]:
+        """
+        Query QuickBooks for items corresponding to the passed conference
+
+        :param conference:
+        :return:
+        """
+        item_names = quick_books_utils.CONFERENCE_TO_LINE_ITEM_NAMES[conference]
+
+        # return Item.choose(item_names, field="Name", qb=self.quickbooks_client)
```
srisainachuri, line 370: commented out code

huxley/invoice_automation/src/module/quick_books_module.py
```diff
@@ -0,0 +1,383 @@
...
+        return self.try_with_refresh(Item.choose, item_names, field="Name", qb=self.quickbooks_client)
+
+    def get_credit_card_processing_fee_item(self) -> Item:
+        if self.credit_card_processing_fee is None:
+            # items = Item.choose([CC_FEE_ITEM_NAME], field="Name", qb=self.quickbooks_client)
```
srisainachuri, line 375: commented out code

srisainachuri (changes requested):
Overall code quality/functionality LGTM!

Major changes requested:
- Error handling so we don't return a non-200 response if the registration was properly serialized but the API call failed for any reason. This ensures advisors registering don't think their registration didn't go through successfully on their end. 
- Non-blocking but we should also implement an easier way to track registrations whose invoices haven't been sent out since having to parse Django logs might be tedious. 

Other minor changes can be implemented in subsequent issues since they don't have a major impact on functionality right now.

skberkeley:
Gotcha, I'll implement a handler which logs an error while still returning the 200 code. I'm down to add a field to the Registration model which indicates whether invoices were sent successfully. @sonikavuyyuru could this field be written to the Huxley data spreadsheet too?

huxley/invoice_automation/src/module/quick_books_module.py
skberkeley: I included this and other commented out code for readability, since it might not be immediately obvious how the syntax of the retry wrapper translates to the actual function call. Would you still want to remove? I can also add something like `# try with retry:` or something similar to make it clear that the comment is showing what the function call w/o retry would look like

srisainachuri (approved):
LGTM!

PR #817, opened 2022-09-27
Update CI Badge in readme

erwang01:
Travis-ci has moved to travis-ci.com


srisainachuri:
I actually don't think we use travis anymore haha, we just use github actions since running out of the free tier, so the travis build status can be removed.

erwang01:
ah okay gotcha. removed.

PR #819, opened 2022-10-04
More verbose logging for invoice automation exceptions

sonikavuyyuru:
Authentication error occurring a while after deployment, so Sora added more verbose logging for invoice automation exceptions


erwang01 (approved):
LGTM if its still needed lol

PR #831, opened 2023-12-19
Update registration sheet

NamanSatish:
Adding the Waitlisted field into the sheet output of registrations.


sonikavuyyuru (approved):
lgtm!

